PR #30, opened 2020-11-02
Simplify proxy list

dufoli:


dufoli:
open a ticket to manage the issue
https://github.com/wildfly/jandex/issues/94

PR #74, opened 2020-12-09
Added support for publishedEndpointUrl attribute of jaxws:endpoint

vhmolinar:
Hello. First of all thanks for your efforts on making quarkus with soap a reality.
So...
I've added support for a new atribute of CXF called **publishedEndpointUrl**.
It is responsible for setting the soapAddress of WSDL and this is well described in the docs:

> The URL that is placed in the address element of the wsdl when the wsdl is retrieved. If not specified, the address listed above is used. This parameter allows setting the "public" URL that may not be the same as the URL the service is deployed on. (for example, the service is behind a proxy of some sort).

http://cxf.apache.org/docs/jax-ws-configuration.html

Please consider this pull request as it is essencial for my project as well I think it would be for many others.

### DIFF
--- a/deployment/src/test/java/io/quarkiverse/cxf/deployment/test/CxfServiceTest.java
+++ b/deployment/src/test/java/io/quarkiverse/cxf/deployment/test/CxfServiceTest.java
@@ -50,6 +50,11 @@ public class CxfServiceTest {
                 .evaluate(doc);
         Assertions.assertEquals("xs:string", val);
 
+        val = xpath
+                .compile(
+                        "/definitions/service/port/address/@location")
+                .evaluate(doc);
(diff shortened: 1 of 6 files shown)


dufoli:
Thanks for this PR. If you can provide a test that will be perfect.
Else it is fine to be merged like that.
Anyway, I am working to move to vertx routing. It will be ready soon and I will need to merge that in my branch.
Many thanks for this ;-)

vhmolinar:
> Thanks for this PR. If you can provide a test that will be perfect.
> Else it is fine to be merged like that.
> Anyway, I am working to move to vertx routing. It will be ready soon and I will need to merge that in my branch.
> Many thanks for this ;-)

Ok. I'll give a look into the tests project

vhmolinar:
@dufoli the test is commited, let me know if you need anything else

PR #81, opened 2020-12-14
Fix issues when using JAXBElement in native mode

bnazare:
This PR fixes multiple issues related to using wrappers with JAXBElement fields in native mode. These issues do not cause problems during compilation but do at runtime.

Here's a rundown of the changes:
1. Added the JAXB API to the list of indexed dependencies. This was necessary as the current code is trying to get `JAXBElement` information from Jandex and without it being indexed all subsequent conditionals fail.
(I'm not sure if I fixed this in the most elegant way, but for now it works.)
2. Fixed a few instances of calls being made to Jandex using the JVM class name format (e.g. "Lmy.class.name;") instead of just the class name.
3. Changed the way "create" methods in the `ObjectFactory` are looked up so that the **type parameter** of `JAXBElement` is used as the method argument instead of `JAXBElement` itself.
4. Fixed an issue with the dynamic byte-code creation when `JAXBElement` is detected. Previously, it caused an error by attempting to access an out-of-scope reference.

### DIFF
--- a/deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
+++ b/deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
@@ -73,6 +73,7 @@ import io.quarkus.deployment.builditem.nativeimage.NativeImageProxyDefinitionBui
 import io.quarkus.deployment.builditem.nativeimage.NativeImageResourceBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.RuntimeInitializedClassBuildItem;
+import io.quarkus.gizmo.AssignableResultHandle;
 import io.quarkus.gizmo.BranchResult;
 import io.quarkus.gizmo.BytecodeCreator;
 import io.quarkus.gizmo.ClassCreator;
@@ -402,7 +403,7 @@ class QuarkusCxfProcessor {
 
     private String createWrapperHelper(ClassOutput classOutput, String pkg, String className,
             MethodDescriptor ctorDescriptor, List<MethodDescriptor> getters,
-            List<MethodDescriptor> setters, IndexView index) {
+            List<MethodDescriptor> setters, List<String> jaxbElementParams, IndexView index) {
         //WrapperClassGenerator
         int count = 1;
         String newClassName = pkg + "." + className + WRAPPER_HELPER_POSTFIX + count;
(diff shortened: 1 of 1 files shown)


deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
```diff
@@ -1332,6 +1343,14 @@ public void build(
         }
     }
 
+    private String extractJAXBElementParam(FieldInfo f) {
+        if (f.type().name().equals(DotName.createSimple(JAXBElement.class.getName()))) {
+            return f.type().asParameterizedType().arguments().get(0).name().toString();
+        } else {
```
bnazare, line 1349: I'm doing this with a simple equality and not a hierarchy analysis because type parameters can come and go with each new sub-class and I don't know how to find the right one so I kept it simple.

Of course, that can always be added later.

dufoli:
wow thanks @bnazare very good point. I have review few stuff. 
point 1: ohh yes good catch previously, I use class loading and it was faling for some people. So yandex is better for that but we need to identified the things to index.
point 2:  not sure of this point because getReturnType must ever been in jvm format
point 3: good catch 
point 4: ok

Thank you very much for this patch.
I really need a unit test for this part. But I focus on cxf pach to throw all and reuse cxf code directly but it take time and in between, I am thanksfull for this help ;-)
I have left some comment.
Code is pretty horrible. I need to split big method to sub methods and even class to make it more readable.

bnazare:
hi @dufoli
On point 2, yes, I think all cases were related to `getReturnType()` returning the JVM format.

On using CXF code directly, I agree, that definitely sounds like a good idea. The scope of JAXB and JAX-WS combined is too large to try to (kinda) reimplement inside a Quarkus extension.

And finally, on tests, the best possible scenario here would be to obtain a WSDL for a publicly available service, generate the JAXB classes during the build and then make a call to one or two operations during the integration tests. IMHO, this would represent the most common scenario for the client side of the extension. But of course, making arbitrary SOAP calls to a service on every build might make its owner a bit angry ....

dufoli:
you can simulate a web service answer as I do with restassure which return the xml of the service.

bnazare:
That sounds interesting, can you point me at that code?

dufoli:
https://github.com/quarkiverse/quarkiverse-cxf/blob/914e2f0b62bb3b1c94571c1f996c05a89919ab1a/deployment/src/test/java/io/quarkiverse/cxf/deployment/test/CxfServiceTest.java#L63

dufoli:
else you can just provide it with a path annotation and post annotation on method and it return the XML of SOAP. So you will use rest api server to test the soap client.
https://quarkus.io/guides/rest-json

Other solution is to use a server and a client of quarkus extension.

dufoli:
not needed anymore with the new code generation which use the cxf one;

PR #87, opened 2020-12-30
start to remove class generation on extension side

### LINKED ISSUE #47: refactoring: class generation injection in cxf for jvm
currently, we use the class generated and inject it through substitution.
Goal here is to ask cxf to modify lib to allow injection of class on jvm part. 
I have open a ticket on cxf jira for that : 
https://issues.apache.org/jira/browse/CXF-8366
next is to push a pull request to them and used the code added.

advantage: speed up boot time

dufoli:
fix #47


Dufgui:
🍾

PR #97, opened 2021-01-11
test WSDL-generated SEI

### LINKED ISSUE #85: Port type name is used in place of the service name
In [this line](https://github.com/quarkiverse/quarkiverse-cxf/blob/8daae3bd3e898c7860025ac32370851e8c775f6d/deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java#L1091) the value of attribute `name` from annotation `@WebService` is taken and then in [this line](https://github.com/quarkiverse/quarkiverse-cxf/blob/8daae3bd3e898c7860025ac32370851e8c775f6d/runtime/src/main/java/io/quarkiverse/cxf/CxfClientProducer.java#L35) that value is used to set the service name. This causes errors whenever port type name and service name don't happen to match.

This happens because, in an unintuitive twist, the `name` attribute actually refers to the port type name and the service name is instead in the `serviceName` attribute. You can confirm this in [the javadoc](https://docs.oracle.com/javase/7/docs/api/javax/jws/WebService.html#name()).

To complicate matters further you can't have the `serviceName` attribute in endpoint interfaces (SEIs) and tools like "wsdl2java" won't generate code with it there. More information [here](https://docs.oracle.com/javase/7/docs/api/javax/jws/WebService.html#serviceName()) and [here](http://cxf.547215.n5.nabble.com/wsdl2java-does-not-generate-serviceName-in-WebService-Annotation-td558502.html#a558503).

If you'd like to reproduce the issue, you can use the following configuration, after generating the appropriate classes from the WSDL:
```properties
quarkus.cxf.endpoint."calculator.asmx".wsdl=http://www.dneonline.com/calculator.asmx?wsdl
quarkus.cxf.endpoint."calculator.asmx".client-endpoint-url=http://www.dneonline.com
quarkus.cxf.endpoint."calculator.asmx".service-interface=org.tempuri.CalculatorSoap
quarkus.cxf.endpoint."calculator.asmx".endpoint-namespace=http://tempuri.org/
quarkus.cxf.endpoint."calculator.asmx".endpoint-name=CalculatorSoap
```

A quick fix could be to add two new configuration parameters for the name and namespace of the service (like you have for the endpoint). A more elaborate fix could make use of classes annotated with `@WebServiceClient` which usually contain this information and are readily generated by tools like "wsdl2java".

bnazare:
As discussed in #85, here's an implementation of a test using a SEI generated from a WSDL. Some notes:
- The WSDL used was taken from an official SoapUI tutorial so I just assumed it was both sufficiently representative and not encumbered by any confidentiality issues. In any case, no calls are being made to the original service, either to get the WSDL or to invoke the service.
- ~~This changeset includes a workaround for #85 (using a bindings file) so it will pass the tests for now. This bindings files should eventually be removed as a way to test that any fix for the issue is working correctly. Please note that this will rename the class `Calculator` back to `CalculatorSoap` so all references will have to be updated.~~ **Removed**
- I don't really like the idea of testing both the server side and the client side at the same time from within Quarkus. I'm concerned that at some point one error in the extension implementation might cancel out when happening on both sides. So, I added a CXF "mini-server" that is started/stopped around the relevant tests and pretends that it is an external service. This "mini-server" does not use Quarkus or this extension. This part of the PR is not essential for the main functionality so you can remove it if you want.
- ~~Currently, none of the wrappers generated uses `JAXBElement` so that part is not being tested yet. I'm currently investigating how to get that part tested and may have something soon.~~ **Implemented**

Now also fixes #85.

### DIFF
--- a/deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
+++ b/deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
@@ -255,11 +255,6 @@ class QuarkusCxfProcessor {
                     } else {
                         wsName = implementor;
                     }
-                    AnnotationInstance webserviceClient = wsClass.classAnnotation(WEBSERVICE_CLIENT);
-                    if (webserviceClient != null) {
-                        wsName = webserviceClient.value("name").asString();
-                        wsNamespace = webserviceClient.value("targetNamespace").asString();
-                    }
                     additionalBeans.produce(AdditionalBeanBuildItem.unremovableOf(implementor));
                     AnnotationInstance bindingType = wsClass.classAnnotation(BINDING_TYPE_ANNOTATION);
                     if (bindingType != null) {
--- a/deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
+++ b/deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
@@ -241,6 +241,12 @@ class QuarkusCxfProcessor {
                 generateCxfClientProducer(generatedBeans, seiClientproducerClassName, sei);
                 unremovableBeans.produce(new UnremovableBeanBuildItem(
                         new UnremovableBeanBuildItem.BeanClassNameExclusion(seiClientproducerClassName)));
+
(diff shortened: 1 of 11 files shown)


dufoli:
Thanks so much. I have added support for WebServiceClient annotation. So you can remove workaround

PR #101, opened 2021-01-14
fix unchecked casts

bnazare:
I noticed there were some unchecked cast warnings during compilation so I made some changes that:
- remove the warnings.
- add some hard checks.
- tidy up the surrounding code.


dufoli:
@bnazare your patch was very good. I am thinking to give you commit access to project.

bnazare:
hi @dufoli, I'm glad you found the code to be of such great quality and I would welcome your proposal if not for a little issue. You see, my job tends to make some sudden and unexpected turns which means I will at some point in the medium-term suddenly stop working with this stuff for some months. For that reason, I think it may be better if we keep the current model of "episodic" (for lack of a better word :-) ) collaboration.

Do expect a few more PRs in the near future, though ;).

dufoli:
Ok. Commit access do not make contributions mandatory. Look to @Dufgui. He has start the project and we talk about it but he do not contribute a lot. And if you need review, you can create a pull request but not accept it... Anyway, it was just a proposal. Thanks for all your work.

Issue #76, opened 2020-12-10
Cannot build native image: NoClassDefFoundError


vhmolinar:
Hello. I am getting lots of NoClassDefFoundError when i try to build native images using graalvm 11.

**Build command:**
`./gradlew build -Dquarkus.package.type=native -Dquarkus.native.container-build=true`

**Error log:**
https://gist.github.com/vhmolinar/3f9dfa9836b2006699d4c19bf4565a51

**Version:**
I tried with 0.1.0 and master. Not 0.1.1 as for some reason the service wasnt ever being deployed.

If necessary I can provide a project with example, but the simplest test will give the same errors above.

dufoli:
Thanks for this report.

it is relate to fastinfoset... which is load by woodstox.
Can you post the version of cxf you used?
mvn dependency tree will be perfect to unsertstand which version is linked to this issue.
Because I know that 3.2 of cxf migrate to woodstox 5.0
version of java and graal will be perfect too.

dufoli:
it was ever repported in #54 
bug come from : https://github.com/oracle/graal/issues/1721
 bug in graal which is ever fixed on Graal 20.3 but quarkus is not based on it currently.
I have added my own patch but not sure it is working.
So just have to wait the switch to 20.3.
https://github.com/quarkusio/quarkus/pull/13644

dufoli:
here is current build warning link to this 
Warning: class initialization of class com.sun.xml.bind.v2.runtime.output.FastInfosetStreamWriterOutput failed with exception java.lang.NoClassDefFoundError: org/jvnet/fastinfoset/VocabularyApplicationData. This class will be initialized at run time because option --allow-incomplete-classpath is used for image building. Use the option --initialize-at-run-time=com.sun.xml.bind.v2.runtime.output.FastInfosetStreamWriterOutput to explicitly request delayed initialization of this class.

dufoli:
ok so I can wait few day for release of quarkus with graal 20.3 else 1.11.0.CR1

dufoli:
I have update to 1.11 so it must has been fixed now.

PR #105, opened 2021-01-15
fix native build without "--allow-incomplete-classpath" argument

bnazare:
This PR removes the "--allow-incomplete-classpath" argument from the native build of the integration tests and fixes the extension so it builds without it.

I'm not a great fan of the "--allow-incomplete-classpath" option outside of debug tasks as I believe it has the potential to sweep both know and unknown issues under the rug. These issue then become visible only at runtime (after a prolongued native compilation) and usually can only be seen/fixed one at a time. In this case it is even worse as it is also forcing the users of the extension to enable that option thus causing the same problems for them.

As far as stability goes, this is basically the same code that I have used for a few months in my company's own application and it has worked solidly so far.

It should also be "neutral" in the sense that if the missing JARs are added, the normal behaviour should resume and if they're not then both JVM and native versions will fail at runtime (though the native version might fail slightly earlier).


dufoli:
I was thinking to remove it too.
And some bugs seems related to it. org/jvnet/fastinfoset/VocabularyApplicationData has been reported in #76. Great!

PR #111, opened 2021-01-25
 code cleaned up of runtime module

bnazare:
Similar to #108 , a bit of code clean-up in module "runtime". Namely:
- Removed unused code
- Added final modifiers where applicable
- Sprinkled in some lambda functions in place of anonymous classes
- Fixed strings comparisons
- Added a metric ton of `@Override` annotations
- Replaced `Exception` in catch clauses with more specific exceptions
- Fixed some JavaDocs

Sorry if this touches too many files, at first I meant to split it into several PRs. However, I might have to leave this stuff for a while so I'm dumping all I've got for now :).

### DIFF
--- a/runtime/src/main/java/io/quarkiverse/cxf/CXFClientInfos.java
+++ b/runtime/src/main/java/io/quarkiverse/cxf/CXFClientInfos.java
@@ -1,5 +1,7 @@
 package io.quarkiverse.cxf;
 
+import java.util.Objects;
+
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.inject.Any;
 import javax.enterprise.inject.Instance;
@@ -28,7 +30,7 @@ public class CXFClientInfos {
 
     public CXFClientInfo getClientInfoBySei(String sei) {
         for (CXFClientInfo clientInfo : cxfClientInfos) {
-            if (clientInfo.getSei() == sei) {
+            if (Objects.equals(clientInfo.getSei(), sei)) {
                 return clientInfo;
             }
         }
--- a/runtime/src/main/java/io/quarkiverse/cxf/CXFQuarkusServlet.java
+++ b/runtime/src/main/java/io/quarkiverse/cxf/CXFQuarkusServlet.java
@@ -6,7 +6,6 @@ import java.util.List;
 import javax.inject.Singleton;
 import javax.servlet.ServletConfig;
 
-import org.apache.cxf.Bus;
 import org.apache.cxf.BusFactory;
 import org.apache.cxf.common.spi.GeneratedNamespaceClassLoader;
 import org.apache.cxf.common.spi.NamespaceClassCreator;
@@ -35,6 +34,7 @@ import io.quarkus.arc.Unremovable;
 public class CXFQuarkusServlet extends CXFNonSpringServlet {
 
     private static final Logger LOGGER = Logger.getLogger(CXFQuarkusServlet.class);
+    private static final long serialVersionUID = 1371666057918777501L;
 
     public CXFQuarkusServlet() {
     }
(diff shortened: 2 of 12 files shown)


runtime/src/main/java/io/quarkiverse/cxf/CXFClientInfos.java
```diff
@@ -28,7 +30,7 @@ public static CXFClientInfo fromSei(String sei) {
 
     public CXFClientInfo getClientInfoBySei(String sei) {
         for (CXFClientInfo clientInfo : cxfClientInfos) {
-            if (clientInfo.getSei() == sei) {
+            if (Objects.equals(clientInfo.getSei(), sei)) {
```
bnazare, line 33: The original code yielded true when both strings were null, but I'm not sure if that was actually a requirement. This code maintains the same behaviour but let me know if I should change it to a simple `String.equals()`.

runtime/src/main/java/io/quarkiverse/cxf/CXFQuarkusServlet.java
```diff
@@ -57,7 +57,7 @@ private Object getInstance(String className) {
         Class<?> classObj = loadClass(className);
         try {
             return classObj.getConstructor().newInstance();
-        } catch (Exception e) {
+        } catch (ReflectiveOperationException | RuntimeException e) {
```
bnazare, line 60: The original code was catching `Exception` which means it was implicitly catching `RuntimeException`s too. For that reason I kept `RuntimeException` in the revised catch clause. This happens in a few other places too.
Ideally, catch clauses with `RuntimeException`s should be removed except where strictly necessary, but I didn't want to make that call at this point.

runtime/src/main/java/io/quarkiverse/cxf/transport/VertxServletOutputStream.java
```diff
@@ -20,7 +20,7 @@
 
 public class VertxServletOutputStream extends ServletOutputStream {
 
-    private HttpServerRequest request;
+    private final HttpServerRequest request;
     protected HttpServerResponse response;
     private ByteBuf pooledBuffer;
     private long written;
```
bnazare, line 26: Fields `written` and `finished` are written to but never read which means they can be removed. Let me know if you want me to do this.

bnazare, line 26: Fields `written` and `finished` are written to but never read which means they can be removed. Let me know if you want me to do this.

runtime/src/main/java/io/quarkiverse/cxf/CXFQuarkusServlet.java
```diff
@@ -35,6 +34,7 @@
 public class CXFQuarkusServlet extends CXFNonSpringServlet {
 
     private static final Logger LOGGER = Logger.getLogger(CXFQuarkusServlet.class);
+    private static final long serialVersionUID = 1371666057918777501L;
```
dufoli, line 37: I think it was added by mistake...

runtime/src/main/java/io/quarkiverse/cxf/CXFQuarkusServlet.java
```diff
@@ -35,6 +34,7 @@
 public class CXFQuarkusServlet extends CXFNonSpringServlet {
```
dufoli, line 34: indeed this class can be totaly removed because I use vertx route now ;-)

dufoli (commented):
sorry I merge it but servlet can be totaly removed indeed and you add a serialVersionUID field

dufoli:
@bnazare I forget to thank you for this patch ;-)

PR #128, opened 2021-02-12
Allow to specify client url as an absolute url

jakubpietrzak96:
In my opinion current approach when specifying the client url is a bit confusing.

```
quarkus.cxf.endpoint."/foo".client-endpoint-url=http://localhost:8081/soap/greeting
```
This configuration requires that the part `/foo` is constant. And if that suffix changes, it is necessary to modify all config entries to accommodate that change.

I think the ability to specify the absolute url of the client service could be useful in some cases - especially if that url needs to be dynamically set e.g via an environment variable.

Looking forward to your thoughts on that (or maybe I missed some obvious solution with the current impelmentation?).

### DIFF
--- a/runtime/src/main/java/io/quarkiverse/cxf/CXFRecorder.java
+++ b/runtime/src/main/java/io/quarkiverse/cxf/CXFRecorder.java
@@ -30,7 +30,8 @@ public class CXFRecorder {
                 String cfgSei = cxfEndPointConfig.serviceInterface.get();
                 if (cfgSei.equals(sei)) {
                     String endpointAddress = cxfEndPointConfig.clientEndpointUrl.orElse("http://localhost:8080");
-                    if (!relativePath.equals("/") && !relativePath.equals("")) {
+                    boolean clientEndpointAbsoluteUrl = cxfEndPointConfig.clientEndpointAbsoluteUrl.orElse(false);
+                    if (!clientEndpointAbsoluteUrl && !relativePath.equals("/") && !relativePath.equals("")) {
                         endpointAddress = endpointAddress.endsWith("/")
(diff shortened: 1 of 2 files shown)


dufoli:
can you open an issue, and we can talk with @shumonsharif  @Dufgui @bnazare or any other of community about better solution.  I want to avoid new boolean config if possible...

runtime/src/main/java/io/quarkiverse/cxf/CXFRecorder.java
```diff
@@ -30,7 +30,8 @@
                 String cfgSei = cxfEndPointConfig.serviceInterface.get();
                 if (cfgSei.equals(sei)) {
                     String endpointAddress = cxfEndPointConfig.clientEndpointUrl.orElse("http://localhost:8080");
-                    if (!relativePath.equals("/") && !relativePath.equals("")) {
+                    boolean clientEndpointAbsoluteUrl = cxfEndPointConfig.clientEndpointAbsoluteUrl.orElse(false);
```
dufoli, line 33: maybe if start with / it mean it is relative url and if start with scheme (http) it is absolute one ???

shumonsharif:
I'm in complete agreement with you @dufoli - we should not be introducing a new config, if at all possible.

@jakubpietrzak96 Would you kindly share some more details on what you are trying to achieve, and why it can't be handled with Quarkus' property expressions and environment variables?
https://quarkus.io/guides/config-reference#combine-property-env-var

shumonsharif:
An additional note - I haven't done this before, but if you must have a need to make `/foo` dynamic, I believe a much better approach would be through:  https://smallrye.io/docs/smallrye-config/interceptors/interceptors.html#relocate-interceptor

jakubpietrzak96:
Closing as agreed in #131

Issue #58, opened 2020-11-20
Unable to use Camel Quarkus with this extension


Nillkki:
**Describe the bug**
Adding this extension to POM breaks Quarkus startup(?) when also using Camel Quarkus.

**Expected behavior**
This extension should work with Camel Quarkus extensions. Running `mvn verify` with the example should result to success.

**Actual behavior**
Running Quarkus/test gives the following error:
```java.lang.RuntimeException: 
java.lang.RuntimeException: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
        [error]: Build step io.quarkus.jaxb.deployment.JaxbProcessor#processAnnotationsAndIndexFiles threw an exception: java.lang.RuntimeException: java.lang.ClassNotFoundException: //org/apache/camel/model/AggregateDefinition
        at io.quarkus.jaxb.deployment.JaxbProcessor.handleJaxbFile(JaxbProcessor.java:232)
        at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
        at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:177)
        at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:177)
        at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
        at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
        at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133)
        at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
        at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
        at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:497)
[...]

PR #130, opened 2021-02-13
Fix for issue #58

shumonsharif:
@dufoli I believe thanks to your more recent updates, combined with this small change, we should be able to close this issue?  Let me know your thoughts!


dufoli:
Have you check that it has fixed it?

shumonsharif:
Yes @dufoli all my testing has been successful, including with using Camel Quarkus and this extension in the same project.  Just wanted to make sure with you that removing this line here has to major impact I'm unaware of; I couldn't think of a good reason it would be required.

```java
forceJaxb.produce(new JaxbFileRootBuildItem("."));
```

Issue #92, opened 2021-01-07
userPrincipal is null in WebserviceContext


neXus1987:
Hi,

after upgrading from 0.2 to 0.3 the following principle is null:

 	@Resource
	WebServiceContext context;
....
_logger.error("auth: " + context.getUserPrincipal());

log:
auth: null

the IndentityProvider will have a principle, maybe this is regarding the update to vert.x?

Thanks
Phil

dufoli:
Do you use  resource in a webservice class. Maybe it is because it is not a bean....
I do not understand how it interact with webservice ??? Can you give more context code to reproduce...

neXus1987:
OK:
Interface:
```
import javax.jws.WebMethod;
import javax.jws.WebParam;
import javax.jws.WebResult;
import javax.jws.WebService;

@WebService(name = "Test")
public interface Test {

	@WebMethod
	@WebResult
	public String test(@WebParam(name = "val") String val);

}
```
Implementation:
```
import javax.annotation.Resource;
import javax.xml.ws.BindingType;
import javax.xml.ws.WebServiceContext;
import javax.xml.ws.soap.SOAPBinding;

import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

@BindingType(value = SOAPBinding.SOAP12HTTP_BINDING)
public class TestEndpoint implements Test {
	@Resource
	WebServiceContext context;
	private static final Logger _logger = LoggerFactory.getLogger(TestEndpoint.class);

	@Override
	public String test(String val) {


		_logger.info("test:" + context.getUserPrincipal());
		return val;
	}

}
```
and with Version 0.2 the userPrinciple was filled and with version 0.3 it is null. The indentityProvider logic will be executed and fills also the userprinciple, but i guess that it were 2 different WebServiceContext.

dufoli:
it must has been fixed by commit of #52 can you test again ?

neXus1987:
With the current version of 0.4.0-SNAPSHOT the injected Bean is null (had added a new service Bean for testing) and the principle is also still null, the context object is not null (as with 0.3):

Endpoint:
```
import javax.annotation.Resource;
import javax.inject.Inject;
import javax.xml.ws.BindingType;
import javax.xml.ws.WebServiceContext;
import javax.xml.ws.soap.SOAPBinding;

import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

import de.hellwach.service.HWMsvV1;

[...]

Issue #109, opened 2021-01-19
@RolesAllowed() not working


wh81752:
I have a HTTP webservice providing REST and SOAP endpoints:

Here's the REST service:
```
@RequestScoped
public class FruitResource {

    @GET
    @Path("/list")
    @RolesAllowed("USER")
    public List<String> list() {
         final SecurityIdentity securityIdentity =CDI.current()
                 .select(SecurityIdentity.class)
                 .get();
        System.out.println(securityIdentity.hasRole("USER"));
        return Arrays.asList( .. );
    }
}
```

Here's the "same" service as SOAP:

```
@WebService(endpointInterface = "org.acme.cxf.FruitWebService")
public class FruitWebServiceImpl implements FruitWebService {

    @Override
    @RolesAllowed("USER")
    public List<String> list() {
        final SecurityIdentity securityIdentity =CDI.current()
                 .select(SecurityIdentity.class)
                 .get();
        System.out.println(securityIdentity.hasRole("USER"));
        return Arrays.asList( .. );
    }
```

This does not work at all.  I'm getting a `javax.enterprise.context.ContextNotActiveException: null` in `io.quarkus.arc.impl.ClientProxies#getDelegate` while all works fine using the REST EP.

My configuration is rather simple:
```
quarkus.http.auth.basic=true
quarkus.security.users.embedded.enabled=true
quarkus.security.users.embedded.plain-text=true
quarkus.security.users.embedded.users.x=x
quarkus.security.users.embedded.roles.x=USER     
quarkus.resteasy.path=/rest
quarkus.cxf.path=/soap
quarkus.cxf.endpoint."/fruit".implementor=org.acme.cxf.FruitWebServiceImpl
```

Not sure whether this is a bug. However, I somehow expect that configured (http) security works regardless of the service implementation.

dufoli:
have you added quarkus security to pom ?

wh81752:
Let me compile an example to support my claim...

wh81752:
Compiled a sample project, available here:

https://github.com/wh81752/cxfdemo-quarkus 

Contains a simple helloworld/greeter service provides as CXF Soap and REST. I've added simple security based on basic auth and user-properties for demonstration purposes.

The methods in question is 
```
    // HelloWorld.java
    @WebMethod()
    @XmlElement(required = true)
    @RolesAllowed({ "APPUSER" })
    String securedHiToUser(@XmlElement(required = true) @WebParam(name = "user") User user);
[...]

Issue #106, opened 2021-01-18
@RequestScoped WebService throws ContextNotActiveException at runtime


wh81752:
http://cxf.apache.org/docs/jaxrs-services-configuration.html states:

> When service classes and providers are registered this way, the default life-cycle is 'singleton'. You can override it by setting a "jaxrs.scope" parameter with the value of 'prototype' (equivalent to per-request).

Trying 

```
@RequestScoped
@WebService(endpointInterface = "org.acme.cxf.FruitWebService")
public class FruitWebServiceImpl implements FruitWebService {

  @Override
   List<Fruit> list() { .. }
}
```

does not work. I'm getting 

```
Caused by: javax.enterprise.context.ContextNotActiveException
        at io.quarkus.arc.impl.ClientProxies.getDelegate(ClientProxies.java:40)
        at FruitWebService_ClientProxy.arc$delegate(FruitWebService_ClientProxy.zig:xx)
        at FruitWebService_ClientProxy.list(FruitWebService_ClientProxy.zig:xx)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at org.apache.cxf.service.invoker.AbstractInvoker.performInvocation(AbstractInvoker.java:179)
```

Background: The need for a request scoped webservice is that I want to inject a request scoped dependency, especially a a request scoped entity manager.

coelho-extremosistemas:
I am trying to Inject others @RequestScoped instances, into a @Webservice class, return null values.

ps: Change to release 0.4.0-SNAPSHOT
ps2: I need to inject datasource instance

wh81752:
> I am trying to Inject others @RequestScoped instances, into a @webservice class, return null values.

Are you trying to do something like this?
```
/* @ApplicationScoped or just absent */
@WebService(endpointInterface = "org.acme.cxf.FruitWebService")
public class FruitWebServiceImpl implements FruitWebService {

   @Inject
   MyRequestScopedClass  mrsc;
   ..
}
```

[...]

PR #144, opened 2021-02-20
Allow access to SecurityIdentity

shumonsharif:
This should address issue #92 - @dufoli Can you please review and provide any feedback.


shumonsharif:
I believe both issues described in #106 should also be fixed with this PR.

shumonsharif:
Just realized that issue #109 should also be fixed with this same PR.

dufoli:
WOW @shumonsharif  great patch. This is the one I plan to do in #109 . PERFECT!

PR #146, opened 2021-02-21
Improved security and error handling

shumonsharif:
@dufoli Some small modifications here to improve the security and error handling ...


shumonsharif:
@dufoli Please hold off on merging this ... I need to fix a couple more things.

dufoli:
And squash all commits too

shumonsharif:
Squashed the commits, ready for review and merge.  I was mistakenly trying to address one issue, but realized the behavior is already exactly as it's supposed to be.  If the check for `@RolesAllowed` fails, an HTTP status code of 500 is returned with the SoapFault without any meaningful fault message .. this can probably be handled better by the end users with an interceptor.

dufoli:
@shumonsharif maybe we need exception mapper as done on resteasy extension...

dufoli:
@shumonsharif I wait some feedback this week then I will provide the release because support of interceptor and security are huge step forward!

shumonsharif:
Yes, agree @dufoli I think ExceptionMapper route may be a good option, will take a look!

PR #149, opened 2021-02-22
release 0.5: security, interceptors and live coding, last graal version

dufoli:


dufoli:
@shumonsharif  we have feedback that it s work perfectly. If you are ok, we can release because it contain many new things.
I plan to wait a week but when I see comment of user to confirm that it works. I am tempted to do it quicker.
wdyt ?

shumonsharif:
@dufoli I completely agree ... some awesome features went into this version, I think it's a good decision to release!  A big kudos to you, and to everyone using the extension as well for all their support and feedback!!  I look forward to where this takes us!!  :)

PR #152, opened 2021-02-24
Documentation updates

shumonsharif:
@dufoli Please do a quick review in case you have any feedback.  Otherwise once merged, we can submit a PR to https://github.com/quarkiverse/quarkiverse-docs/blob/master/antora-playbook.yml


dufoli:
Sounds good to me. We have a lot to add later
Code generation with maven plugin
Main method instead of container managed 
A sentence about corba
Cient injection 
Cxf Async
Wsdl
Literal, document, bare, raw mode
ServiceName and port name 
WebServiceClient annotation
Link to cxf doc 
Fastxml
Performance
...

PR #164, opened 2021-03-01
Opentracing feature & interceptors, compatible with quarkus

OpenGuidou:
Quarkus and cxf are implementing two different versions of opentracing api:
0.31.0 for quarkus and 0.33.0 for cxf.

I am here adapting the code of cxf-integration-tracing-opentracing so that it complies with the 0.31.0.
It can then be removed when quarkus is aligned to 0.33.0.

The change compared to the original code is only in io.quarkiverse.cxf.tracing.AbstractOpenTracingProvider.startTraceSpan(Map<String, List<String>>, URI, String), using the HeaderExtractAdapter instead of the TextMapAdapter.

### DIFF
--- a/runtime/pom.xml
+++ b/runtime/pom.xml
@@ -66,6 +66,14 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>io.opentracing</groupId>
+            <artifactId>opentracing-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.opentracing</groupId>
+            <artifactId>opentracing-util</artifactId>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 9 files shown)


dufoli:
thanks but I prefer that you open PR to quarkus directly

here is maintainer FYI:

SmallRye OpenTracing | Pavol Loffay, Ladislav Thon
-- | --

shumonsharif:
To add to @dufoli 's suggestion - it should theoretically be possible to bundle the correct version of `cxf-integration-tracing-opentracing` (which would be compliant with 0.31.0) into your application, and just [use the relevant features and/or interceptors](https://cwiki.apache.org/confluence/display/CXF20DOC/Using+OpenTracing#UsingOpenTracing-DistributedTracingwithOpenTracingandJAX-WSsupport.1) directly.

OpenGuidou:
Thanks for the comments.
I actually wanted to avoid runtime issues by having this project depending on one cxf version and cxf-integration-tracing-opentracing on another.
Anyway this was just a proposal and I understand if the prefered move is to migrate quarkus to the latest opentracing api version.. if that happens one day 😁

dufoli:
just harace maintainer of open tracing extension and open PR 😁😁😁😁😁

Issue #147, opened 2021-02-21
Documentation 


dufoli:
Provide documentation in quarkiverse doc
Add link to it in readme. MD. 
Document process for client, request scoped, server process, add interceptor, security support, native build,...

shumonsharif:
Nice!!!  I always wondered why there wasn't a Quarkiverse website ... glad to see there is one.  Happy to collaborate with you on this in a branch, let me know!

dufoli:
https://quarkiverse.github.io/quarkiverse-docs/quarkus-cxf/dev/index.html

shumonsharif:
Hey @dufoli I don't believe Antora uses or depends on maven.  So we shouldn't actually be adding any pom.xml or module changes to parent pom.  I believe the way quarkiverse-docs hooks in and produces the website is via [this](https://github.com/quarkiverse/quarkiverse-docs/blob/master/antora-playbook.yml), which we'll need to submit a PR for.  Also, the branch I committed yesterday already has a docs folder with similar structure and the content we currently have on our existing README.md, feel free to commit directly into the branch while we finalize what the docs should look like.  I'm copying @gastaldi to confirm this.

gastaldi:
That's right, the `docs` module will not be built on maven, but by Antora. That's the process we're adopting for now.

dufoli:
I have edit my comment to reflect your comment. Thanks ;-)

dufoli:
I have update description with structure. @shumonsharif  does it sounds ok ?

shumonsharif:
Looks perfect, thanks @dufoli!!

dufoli:
We have a lot to add later:
Code generation with maven plugin
Main method instead of container managed
A sentence about corba
Cient injection
Cxf Async
Wsdl
Literal, document, bare, raw mode
ServiceName and port name
WebServiceClient annotation
Link to cxf doc
Fastxml
Performance
...

shumonsharif:
@dufoli that's a great list. I think many of these would be perfect candidates for an example codebase that demonstrates everything in action.

It seems the PR for #150 got closed for now, because we may need to bundle the codestarts as part of the extensions ... once that's available, we should be able to capture a lot of this as code examples.  Of course, we can document too ... what are your thoughts?

omurthy:
You have added your own web server and client to the same project. I think they should be separate.

Can you add a sample WSDL-consuming tutorial? 
[...]

PR #169, opened 2021-03-03
Updated Quarkiverse docs

shumonsharif:
Hey @dufoli this splits out the docs into a more manageable set of docs (instead of the gigantic one we had before).  I also added a few sections in the index page that we haven't tracked, such as contract vs code first, sample code and issues and support sections.


shumonsharif:
Not exactly the same structure you proposed here: https://github.com/quarkiverse/quarkus-cxf/issues/147#issuecomment-783203096 but it is pretty close :)

PR #181, opened 2021-03-26
PR for configuring clients, especially client variants.

wh81752:
Dear Quarkus-CXF Maintainers,

pls find here a PR for a different approach on configuring of CXF clients:

```
    @Inject
    @CXF(config = "featured-foo")                      // pick a dedicated client configuration
    GreetingClientWebService featuredWS; 

    featuredWS.ping("hello");                               // no change in using a proxy client (of course).
```

The annotation CXF is optional. If absent configuration properties are taken from the first client configuration matching the SEI, otherwise some meaningful (?) config values are used.

The configuration is almost as before. However, I introduced prefix "quarkus.cxf.client" to make it hopefully less confusing whether a service is published or whether a client is configured. Here's an example:

```
#   @Inject @CXF(config="foo") GreetingClientWebService myclient;
quarkus.cxf.client."foo".client-endpoint-url=http://localhost:8081/soap/greeting
quarkus.cxf.client."foo".service-interface=io.quarkiverse.it.cxf.GreetingClientWebService

#   @Inject @CXF(config="featured-foo") GreetingClientWebService myclient;
quarkus.cxf.client."featured-foo".client-endpoint-url=http://localhost:8081/soap/greeting
quarkus.cxf.client."featured-foo".service-interface=io.quarkiverse.it.cxf.GreetingClientWebService
quarkus.cxf.client."featured-foo".out-interceptors=io.quarkiverse.it.cxf.MyCxfInterceptor
```

Please be aware that the configuration key has no meaning any longer. It is nothing more than a logical grouping. Before the key got reused as relative path. The problem with this approach is, that the final client URL is not available by grepping the source code.

There are least two problems though:

1. The client proxy for  SEI GreetingWebService appears not to be produced via CxfClientProxy. I suspect that doing so will endup in a bean conflict.
2. If @CXF is absent then one of the matching client configurations is applied. This could be "featured-foo" in the example above. There should  be a more controlled way which configuration is the default for a given SEI. Perhaps by using an extra attribute ".default = true".


dufoli:
thanks for PR, I understand the use case, you want to address. 
but why not used @ named annotation instead of a new one ?

wh81752:
> but why not used @ named annotation instead of a new one ?

This is perhaps where you come into the game :-)

```
@Inject
@Named(config = "featured-foo")                      // pick a dedicated client configuration
GreetingClientWebService featuredWS;
```

Here we refer to a bean named "featured-foo". The problem is that all beans are created at "augmentation" time. At that time however I do not have any information about the current configuration. This is at least my understanding - perhaps I'm wrong.

Thus my workaround so far is to call the producer function with the current InjectionPoint as additional information and let the producer function figure out which config to use.

Agreed @CXF is not pretty. If it can be avoided or replaced then fine with me. Right now it is unclear to me how to do so. 

Perhaps https://quarkus.io/guides/cdi-integration#injection_point_transformation is the way to go? It would replace @Named with @CXF internally.

shumonsharif:
Nice work @wh81752.  Just some thoughts in no specific order:

- With introducing a `quarkus.cxf.client` prefix, would it possibly be cleaner to have a `quarkus.cxf.service` prefix as well?  We'd of course need to have this documented properly, as we'd be introducing multiple non-backward compatible changes.
- I'm not sure whether replacing `@Named` with `@CXF` may work well - it may be difficult to distinguish regular `@Named` annotations that shouldn't be replaced.
- In the PR, I see `org.wildfly.common.annotation.Nullable` in multiple places combined with `Optional.ofNullable(foobar)` ... just quickly glancing at it, is the `@Nullable` annotation actually required ... ie. the latter should achieve the same?  If it is indeed redundant, then we can possibly get rid of the `@Nullable` dependency.
- The client docs will need to be updated, would be great if it were part of the PR.

Having said all this, I like the PR overall, and as @dufoli is the main contributor, will leave it to your decision!

wh81752:
> Nice work @wh81752. Just some thoughts in no specific order:

Thx. Let me reply on your comments and .. this PR is by no means final. I'm happy to discuss, extend, improve and so on.

> The client docs will need to be updated, would be great if it were part of the PR.
Good point, going to add docu as well.

> I'm not sure whether replacing @Named with @CXF may work well - it may be difficult to distinguish regular @Named annotations that shouldn't be replaced.

I'm experimenting with that replacement right now.

Btw: Perhaps the name @CXF should be replaced with something more appropriate. Didn't think much about it. Suggestions are welcome.


> just quickly glancing at it, is the @Nullable annotation actually required

Technically, @Nullable is not required. It's rather part of my own programming convention:

1. If a parameter is annotated with @Nullable then it may be null, otherwise it is *never* null. 
2. @Nullable should be reduced to a minimum
3. Avoid Optional<T> in (function/method) signatures

Following this style results in much more readable code then using Optional<T>. However, since this is my personal view I have little problem to change if required.

> With introducing a quarkus.cxf.client prefix, would it possibly be cleaner to have a quarkus.cxf.service prefix as well?

My main headaches were which of the props are going to be used for configuring a client and which are used for the service to be published. That's why I eventually introduced that new "client" configuration branch.

Although "endpoint" is a quite generic term, I have no problem with that name. I guess it's just a matter of documentation.

wh81752:
Dear All,

> The client docs will need to be updated, would be great if it were part of the PR.

Please find updated client docu as well (btw: how to generate and view docu locally via Maven??).

The last update of this PR addresses also issue #182. Let me shortly summarize here how it now looks like (or pls read docu) with an example:

```
@Application    // or any other CDI scope
public class MySoapClient {

    private boolean logging = ..

    @Inject
    @CXFClient("my-featured-fruitservice-client")
    FruitWebService featuredClientService;                   // (A)

    @Inject
    @CXFClient
    FruitWebService clientService;                                  // (B)

    @Inject
    FruitWebService impl;                                                 // (C)

    @Inject
    @CXFClient("my-featured-fruitservice-client")
    CXFClientInfo featuredClientInfo;                              // (D)

    @Inject
    @CXFClient
    CXFClientInfo clientInfo;                                             // (E)

    public int getCount() {
        System.out.println("using ep: " + this.serviceUrl()); 
        return (logging ? featuredClientService : clientService).count();
    }

   public String serviceUrl() {
        return (logging ? featuredClientInfo : clientInfo).getEndpointAddress();
   }
}
```

Notice that I changed `@CXF` into `@CXFClient` to have a notation that fits more exactly with the annotation's purpose.

Then  `@CXFClient` can also be used to inject meta client information as shown in `(D)` and `(E)`. This can be especially useful when testing clients, i.e. is my client really targeting the webservice I want to target? Or checking that interceptor or feature is really plugged in?

Finally I changed the generated producer classes to not any longer produce a value for injection points like `(C)`. This addresses issue #182. Now the service implementation is injected if around, otherwise CDI complains. Technically only the latter got changed. Before, our generated producer kicked in by injecting a client proxy. The former always worked because quarkus-cxf marks all classes annotated with `@WebService` automatically as beans.

From a functional point of view I'm rather happy with the current implementation. One thing to change is perhaps that an injected `CxfClientInfo` instance should be immutable by nature.

However, I'm open for comments and critics ..

deployment/src/test/java/io/quarkiverse/cxf/deployment/test/CxfClientTest.java
```diff
@@ -16,15 +21,50 @@
...
+    @Inject
+    @Named("io.quarkiverse.cxf.deployment.test.FruitWebService")
+    CXFClientInfo clientInfo;
+
     @Inject
     FruitWebService clientService;
 
+    @Inject
+    @CXFClient
+    FruitWebService clientProxy;
```
dufoli, line 37: What is the difference between cliebtproxy and clientservice? So why add not needed annotation?

dufoli: Hmm I see that one is proxy but what is goal behind switch to proxy

deployment/src/test/resources/application-cxf-test.properties
```diff
@@ -1,2 +1,9 @@
-quarkus.cxf.endpoint."/fruit".client-endpoint-url=http://localhost:8081/
-quarkus.cxf.endpoint."/fruit".service-interface=io.quarkiverse.cxf.deployment.test.FruitWebService
\ No newline at end of file
+
+
+quarkus.cxf.client."fruitclient".client-endpoint-url=http://localhost:8081/fruit
```
dufoli, line 3: Not happy about loosing compatibility of configuration but I have to think if it is better. Endpoint is more for server so make more sens to split client. And avoid the concatenation or URL with relative path. But still not sure. I have to think to it. Need pros cons

runtime/src/main/java/io/quarkiverse/cxf/CxfClientProducer.java
```diff
@@ -104,34 +138,106 @@ private void addToCols(String className, List<Interceptor<? extends Message>> co
...
+
+    /**
+     * Calculates the client info to use for producing a JAXWS client proxy.
+     *
+     * @param cxfConfig The current configuration
+     * @param ip Meta information about where injection of client proxy takes place
+     * @param meta The default to return
+     * @return not null
+     */
+    static private CXFClientInfo selectorCXFClientInfo(
```
dufoli, line 182: I like the idea to move selector here

dufoli: Question about selector do you handle default VS named client?

dufoli (commented):
I have done a full review. Big patch so it take time to think about pros vs cons. Next time best is open an issue to talk about issue you want to address please. Anyway, pretty good code but lot of line are changed (sometimes for style) and it is hard to see if we introduce bugs... And rétro compatibility issue too.

wh81752:
How to proceed by now? I propose to first resolve minor code changes (imports, formatting) by updating this PR so that they are gone.

Perhaps later on I suggest to add a link or hint to a Quarkus style guide. In addition it would be good to be able formatting rules into Intellij to avoid such changes. Or to update format rules attached to Maven or sharpen the checks when building.

runtime/src/main/java/io/quarkiverse/cxf/CxfClientProducer.java
wh81752: > Question about selector do you handle default VS named client?

Don't understand the question? What is VS standing for?

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
```diff
@@ -1032,4 +1071,164 @@ private String getMappingPath(String path) {
...
+            // create method
+            // >> @Produces
+            // >> @CXFClient
+            // >> {SEI} createService(InjectionPoint ip) { .. }
+
+            String p0class = "javax.enterprise.inject.spi.InjectionPoint";
+            String p1class = "io.quarkiverse.cxf.CXFClientInfo";
+            try (MethodCreator createService = classCreator.getMethodCreator("createService", sei, p0class)) {
+                createService.addAnnotation(Produces.class);
+                createService.addAnnotation(CXFClient.class);
```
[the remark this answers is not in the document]
wh81752: Can you please elaborate on this? Don't get it.

deployment/src/main/java/io/quarkiverse/cxf/deployment/GeneratedBeanRecorder.java
```diff
@@ -0,0 +1,62 @@
...
+import io.quarkus.bootstrap.BootstrapDebug;
+import io.quarkus.gizmo.ClassOutput;
+
+/**
+ * Simple bean recorder.
+ *
+ * <p>
+ * Creates a GeneratedBeanBuildItem out of every class written to interface ClassOutput.
+ */
+public class GeneratedBeanRecorder implements ClassOutput {
```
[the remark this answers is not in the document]
wh81752: Hmm, if already existing then of course it can be replaced.

runtime/src/main/java/io/quarkiverse/cxf/CxfClientProducer.java
dufoli: Versus

dufoli:
For coding guidelines https://github.com/quarkusio/quarkus/blob/main/CONTRIBUTING.md#idea-setup
Maybe plugin can hzlp. I do not remember if I install it
https://plugins.jetbrains.com/plugin/13234-quarkus-tools

dufoli:
About named, here is Mongo processor https://github.com/quarkusio/quarkus/blob/main/extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
Which inject client with named.

dufoli:
Sorry to have so much remark whereas you produce a lot of code. But as maintainer, I have avoid to break things or lost compatibility.

deployment/src/test/java/io/quarkiverse/cxf/deployment/test/CxfClientTest.java
wh81752: This is the difference, addressing issue #182:

```
    @Inject
    @CXFClient
    FruitWebService clientProxy;
```

Here @CXFClient asks to insert unconditionally a CXF proxy client. When you use this annotation, you know for sure that you are talking via a proxy to your backend service. You can also configure your "client" for logging and so on.

Consider the situation before:

```
    @Inject
    SomeWebService client;        // SomeWebService supposed to be a SEI
```

What is getting injected here? You need to investige SomeWebService to know. If there is a impl of SEI in your classpath, then you get no proxy at all - you just get the impl itself injected. Contrary, if no impl is around, you get a proxy injected:

```
    @WebService
    public class SomeWebServiceImpl implements SomeWebService {}      // injects  SomeWebServiceImpl

    vs

    @WebService
    public interface SomeWebService {}                                 // injects Proxy (via Producer).
``

This behaviour is rather confusing im my opinion (see #182).

The new annotation @CXFClient resolves this issue. Once attached, you indicate that you want to have a proxy injected. In addition, you can indicate which particular client configuration you want to have.

Contrary to this you may want to have the service implementation injected:

```
    @Inject
     SomeWebService client;    // CDI error if NO implementation is around.
```

File: deployment/src/test/resources/application-cxf-test.properties (hunk 2 above)
Comment (wh81752) [reply deployment/src/test/resources/application-cxf-test.properties:3]: PROS:

1. The complete URL is given in configuration, there is no need to "concat". This makes it easy for searching, i.e. which config is used talking to a service endpoint.

2. Many different client configurations possible.

File: runtime/src/main/java/io/quarkiverse/cxf/CxfClientProducer.java (hunk 3 above)
Comment (wh81752) [reply runtime/src/main/java/io/quarkiverse/cxf/CxfClientProducer.java:182]: Now I understand what VS means, not sure whether I got the question as such.

`selectorCXFClientInfo()` has three arguments, the current config, the injection point and the default meta information for a client of a given SEI. Basically, if no config applies then the meta information is returned. Otherwise a dedicated client info is returned based on given meta info. So in this sense `selectorCXFClientInfo()` handles a default and named client injection point.

**Comment by wh81752** [comment]
> Sorry to have so much remark whereas you produce a lot of code. But as maintainer, I have avoid to break things or lost compatibility.

No problem. Let's just dicuss the PR and let at the pros and cons. As PR creator I see of course only PROS :-) However, there might be also pretty good cons of course.

File: deployment/src/main/java/io/quarkiverse/cxf/deployment/GeneratedBeanRecorder.java (hunk 5 above)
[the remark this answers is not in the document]
Comment (wh81752) [reply deployment/src/main/java/io/quarkiverse/cxf/deployment/GeneratedBeanRecorder.java:22]: I see - obviously I just re-invented the wheel again :-)

**Comment by wh81752** [comment]
While further testing I noticed that 

```
@Inject
@CXFClient
CxfClientInfo info;
```

doesn't work. It went undiscovered cause only one webservice has been present in `CxfClientTest`. Therefore I added SEI GreetingWebService and two impls (GreetingWebServiceImpl and HelloWebServiceImpl) as well in order to discover potential problems earlier.

**Comment by wh81752** [comment]
I also removed (hopefully) all changes that were purly cosmetic. Also installed  [Eclipse Code Formatter Plugin](https://plugins.jetbrains.com/plugin/6546-eclipse-code-formatter) along with Quarkus `eclipse-format.xml` and `eclipse.importorder` to be inline with styling.

**Comment by dufoli** [comment]
By impl, you mean the impl of sei on server part, right? Why injecting it?
For me server impl is never injected only use by Vertx handler.

**Comment by wh81752** [comment]
> By impl, you mean the impl of sei on server part, right? 

Correct. Impl for now means strictly implementation of webservice sei.

> Why injecting it? For me server impl is never injected only use by Vertx handler.

Well, it just get's injected if you write this

```
@Inject 
FruitWebService client;
```

cause there is a build step that turns every class (or interface) annotated with `@WebService` into a bean. 

The business case is injecting impl for the purpose of testing. However, pretty positive that there are other cases where you want the impl and not the proxy.

**Comment by dufoli** [comment]
Hello, sorry I am a bit inactive at the moment but I hope, I will be more active in few weeks. Anyway, @shumonsharif can you review code too.

**Comment by shumonsharif** [comment]
I will try to take a look this weekend, please stay tuned for a proper review.  Just a quick glance shows there's still quite a bit of cleanup that can and should be done.  From underscores in method names, to renaming variables causing a large number of unnecessary code changes, to even small things like `@author` tags that don't add much value, my main issue with this PR as of now - is that it deviates quite a bit in terms of **consistency with the existing codebase**, and thereby makes it very difficult to focus on the more impactful changes that we should be reviewing.

File: integration-tests/src/test/java/io/quarkiverse/it/cxf/SOAPExceptionInterceptor.java [hunk 6]
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * <p>
+ * This interceptor is intented to be used for testing. Configure a
+ * CXF client with this interceptor and simply test whether there is an
+ * effect, i.e. expect a {@link SOAPFaultException} on sending a message.
+ * </p>
+ *
+ * @author wh81752
+ */
+public class SOAPExceptionInterceptor extends AbstractPhaseInterceptor<Message> {
```
shumonsharif, line 22: I'm not sure this class is getting used as intended.  It appears a WebServiceException is thrown in the relevant test case (test_out_interceptor_present), and not a SOAPFaultException.  Kindly fix, as the tests are broken at the moment.

runtime/src/main/java/io/quarkiverse/cxf/CXFClientData.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Provides runtime metadata for a CXF client.
+ *
+ * <p>
+ * This class contains extracted from a SEI. It contains basic data to
+ * setup a proxy client for a given SEI.
+ * </p>
+ *
+ * @author wh81752
```
shumonsharif, line 17: Kindly remove all author tags in various classes.  There is awesome view of all contributors to a project at the following link:  https://github.com/quarkiverse/quarkus-cxf/graphs/contributors

runtime/src/main/java/io/quarkiverse/cxf/CXFClientData.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+public class CXFClientData {
+    private static final Logger LOGGER = Logger.getLogger(CXFClientData.class);
+    final public List<String> classNames = new ArrayList<>();
+    public String soapBinding;
+    public String sei;
+    public String wsName;
+    public String wsNamespace;
+    public @Nullable String serviceName;
```
shumonsharif, line 26: Kindly remove the dependency on the wildfly-common package with Nullable.  This is mainly to reduce dependencies on packages that are not really required, helps maintainability in the long run, and would be consistent with the existing code base.

runtime/src/main/java/io/quarkiverse/cxf/CXFRecorder.java
```diff
@@ -19,79 +20,22 @@
...
-            }
-            return cfg;
-        };
+    /**
+     * Create CXFClientInfo supplier.
+     * <p>
+     * This method is called once per @WebService *interface*. The idear is to produce a default client config for a
+     * given SEI.
+     */
+    public RuntimeValue<CXFClientInfo> cxfClientInfoSupplier(CXFClientData ws) {
```
shumonsharif, line 31: Kindly change `CXFClientData ws` to say `CXFClientData cxfClientData` as it improves readability and maintainability.

runtime/src/main/java/io/quarkiverse/cxf/CxfClientProducer.java
```diff
@@ -27,26 +33,46 @@
...
+    }
 
-        CXFClientInfo cxfClientInfo = getInfo();
+    /**
+     * The main workhorse producing a CXF client proxy.
+     *
+     * @param info
+     * @return
+     */
+    private Object produceCxfClient(CXFClientInfo info) {
```
shumonsharif, line 67: Kindly change `CXFClientInfo info` to `CXFClientInfo cxfClientInfo` as it improves readability and maintainability, and it's also consistent with existing standards in the code base.

runtime/src/main/java/io/quarkiverse/cxf/CxfConfig.java
```diff
@@ -15,4 +18,18 @@
...
+     * Configure client proxiess.
+     */
+    @ConfigItem(name = "client")
+    public Map<String, CxfClientConfig> clients;
+
+    public boolean isClientPresent(String key) {
+        return Optional.ofNullable(clients).map(m -> m.containsKey(key)).orElse(false);
+    }
+
+    public @Nullable CxfClientConfig getClient(String key) {
```
shumonsharif, line 32: Please remove Nullable as per previous comment.

shumonsharif (changes requested):
I just did a full review.  First off, amazing work @wh81752!!  You put in a tremendous amount of work here, thanks for your contribution.  I think the pros for this PR outweigh the cons, and though it introduces non backward compatible changes, I think we can address via documentation and release notes.  Kindly take a look at the feedback provided, and let me know if any questions.

wh81752:
@shumonsharif - thx for reviewing. I will comment and fix mentioned issues soon. Pls stand by.

dufoli:
and please rebase on master ;-)

deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfWebServiceBuildItem.java
```diff
@@ -74,4 +75,14 @@ public String getImplementor() {
     public boolean IsClient() {
         return isClient;
     }
+
+    public CXFClientData clientData() {
```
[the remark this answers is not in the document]
wh81752: The idea behind CXFClientData is to avoid method calls with a large numer of arguments, especially if those arguments do all have the same type.

In this way, CXFClientData is not intended to be JavaBean but rather a Java Record type (introduced in Java 16).

Perhaps a nicer would be to use the builder pattern I guess.

Now what to do. Shall I change into something like

`cxf.setSoapBinding(this.getSoapBinding());`

i.e. simply introduce setters and getters on `CXFClientData()`?

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
```diff
@@ -1032,4 +1028,131 @@ private String getMappingPath(String path) {
...
+                createService.returnValue(createService.checkCast(cxfClient, sei));
+            }
+        }
+
+        // Eventually let's produce
+        produceUnremovableBean(unremovableBeans, cxfClientProducerClassName);
+    }
+
+    @SafeVarargs
+    static private <T> Set<T> asSet(T... items) {
```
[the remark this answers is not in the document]
wh81752: Fixed. Sidenote: would be good if rules like this are somehow enforced. Perhaps using Maven checkstyle?

deployment/src/test/java/io/quarkiverse/cxf/deployment/test/CxfClientTest.java
```diff
@@ -16,15 +21,34 @@
...
+    FruitWebService serviceImpl;
+
+    @Inject
+    @CXFClient
+    FruitWebService proxyClient;
 
     @Test
-    public void whenCheckingClientInjected() {
-        Assertions.assertNotNull(clientService);
+    public void test_injected_beans() {
```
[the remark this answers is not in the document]
wh81752: Done.

deployment/src/test/resources/application-cxf-test.properties
wh81752: > The main con is see here is losing backward compatibility.

With some additional code it would be possible to soften this up allowing the old config as well. A warning could be given that this config style is subject to be disallowed. Nonetheless, given that version is 0.x, I guess that no-one is  expecting backward compatibility yet. Otherwise please raise your hands NOW!!

integration-tests/src/test/java/io/quarkiverse/it/cxf/SOAPExceptionInterceptor.java
wh81752: Ok, let me re-check this later. It was working fine so far. The idea of this interceptor is just disable any message exchange by  forcing CXF to throw an exception.

runtime/src/main/java/io/quarkiverse/cxf/CXFClientData.java
wh81752: Done.

runtime/src/main/java/io/quarkiverse/cxf/CXFClientInfo.java
```diff
@@ -121,4 +145,101 @@ public String getPassword() {
...
+
+    private CXFClientInfo addFeatures(CxfClientConfig cxfEndPointConfig) {
+        if (cxfEndPointConfig.features.isPresent()) {
+            this.features.addAll(cxfEndPointConfig.features.get());
+        }
+        return this;
+    }
+
+    @Override
+    public String toString() {
```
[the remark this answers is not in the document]
wh81752: Hmm, can a toString() method to any harm? Anyway, removed.

runtime/src/main/java/io/quarkiverse/cxf/CXFClientInfo.java
```diff
@@ -121,4 +145,101 @@ public String getPassword() {
...
+                ", outInterceptors=" + outInterceptors +
+                ", outFaultInterceptors=" + outFaultInterceptors +
+                ", inFaultInterceptors=" + inFaultInterceptors +
+                ", features=" + features +
+                ", classNames=" + classNames +
+                '}';
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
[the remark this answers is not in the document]
wh81752: Same comment as above ;-)

runtime/src/main/java/io/quarkiverse/cxf/CXFClientInfo.java
```diff
@@ -121,4 +145,101 @@ public String getPassword() {
...
+                        outInterceptors,
+                        that.outInterceptors)
+                && Objects.equals(outFaultInterceptors, that.outFaultInterceptors) && Objects.equals(
+                        inFaultInterceptors,
+                        that.inFaultInterceptors)
+                && Objects.equals(features, that.features) && Objects.equals(classNames, that.classNames);
+    }
+
+    @Override
+    public int hashCode() {
```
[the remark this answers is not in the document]
wh81752: Removed.

runtime/src/main/java/io/quarkiverse/cxf/CXFRecorder.java
wh81752: Changed. Altough I believe that if there is jsut one parameter and a very short body, the variable name becomes irrelevant. All that matters is the variable's type ..

runtime/src/main/java/io/quarkiverse/cxf/CxfClientProducer.java
wh81752: Fixed.

runtime/src/main/java/io/quarkiverse/cxf/CxfConfig.java
wh81752: Removed, however ..

I really like the following style in Java:
An argument can only be null if the parameter is annotated with something like `@Nullable`. A `@Nullable` parameter is an exception that should be only used when necessary. I really would like to see `@Nullable`  as Java standard.

wh81752: Done.

runtime/src/main/java/io/quarkiverse/cxf/CxfClientProducer.java
wh81752: Done.

runtime/src/main/java/io/quarkiverse/cxf/CXFRecorder.java
wh81752: Done.

runtime/src/main/java/io/quarkiverse/cxf/CXFClientData.java
wh81752: Done.

runtime/src/main/java/io/quarkiverse/cxf/CXFClientData.java
wh81752: Done.

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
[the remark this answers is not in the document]
wh81752: Done.

deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfWebServiceBuildItem.java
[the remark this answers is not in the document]
wh81752: Let me add ctor then ..

runtime/src/main/java/io/quarkiverse/cxf/CXFClientInfo.java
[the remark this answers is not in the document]
wh81752: Done.

runtime/src/main/java/io/quarkiverse/cxf/CXFClientInfo.java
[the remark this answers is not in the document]
wh81752: Done.

runtime/src/main/java/io/quarkiverse/cxf/CXFClientInfo.java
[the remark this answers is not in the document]
wh81752: Done.

deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfWebServiceBuildItem.java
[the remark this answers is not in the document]
wh81752: Ok, I changed CXFClientData into a plain and serializable (!) POJO.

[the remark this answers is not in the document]
shumonsharif: Hey apologies I could have been more clear on this previously - what I meant is why do we need the clientData method at all in CxfWebServiceBuildItem class?  You should use a CXFClientData constructor wherever this clientData() method is being called, and remove this method all together if possible.

deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfWebServiceBuildItem.java
```diff
@@ -20,8 +19,14 @@
     private final String path;
     private final boolean isClient;
 
-    public CxfWebServiceBuildItem(String path, String sei, String soapBinding, String wsNamespace,
-            String wsName, List<String> classNames, String implementor) {
+    public CxfWebServiceBuildItem(
```
shumonsharif, line 22: Please remove all unnecessary formatting changes throughout the PR ... formatting changes like this make it quite difficult to tell at a later point whether something has really changed or not, especially when trying to identify a potential issue introduced in a PR.  Given the complexity of this PR, I would strongly recommend you fix this everywhere applicable.

shumonsharif (changes requested):
Hey @wh81752 Please take a look at the couple comments I just added!

deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfWebServiceBuildItem.java
wh81752: I see the problem: Disabled RedHat's Intellij Quarkus-Plugin because of issues. Now formatting is broken :-(

wh81752: To my best knowledge: fixed.

integration-tests/src/test/java/io/quarkiverse/it/cxf/SOAPExceptionInterceptor.java
wh81752: Works as expected on my side, no broken test locally.

deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfWebServiceBuildItem.java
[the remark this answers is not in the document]
wh81752: I changed it. Nonetheless, don't see any relevant problem. CxfWebServiceBuildItem simply has all the knowledge of creating such an CXFClientData. Now I moved this method into the processor as static function. Fine with me, but I thing that such boiler-plate code is better located in responsible classes while the processor should focus on business logic.

famod:
Really looking forward to this getting merged and released eventually!

One question: With this, will I be able to have multiple clients for the same interface?
In my project we have multiple "vendors" implementing the same webservice interface (or wsdl so to say) and we have to select one of the clients dynamically.
Something like `Instance<FruitWebService >` would come to mind...
In config we'd then have multiple URLs etc.

wh81752:
> One question: With this, will I be able to have multiple clients for the same interface?

@famod - Yes. This use-case has been anticipated. The way it works this way:

```
@Inject @CXFClient("my-other-fruit-service")
FruitWebService otherFruitService;
```

where `my-other-fruit-service` is an arbitrary configuration key where you upfront configured your alternative service.


> Something like Instance<FruitWebService> would come to mind...

Have not tested this yet. Technically it shoud work, however, I guess you need to figure out to whom you are talking then, i.e. what is the particular config behind an instance?

wh81752:
@shumonsharif - All requested changes are done. Could you review again? Or did I miss something?

shumonsharif:
Hey @wh81752 When I click on the ['Files Changed'](https://github.com/quarkiverse/quarkus-cxf/pull/181/files) tab - I don't see most of the changes I had requested in the initial review, that you had marked resolved?  I know I saw commits for them from you ... just not sure why the PR is not getting updated?

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
```diff
@@ -236,12 +256,23 @@ public void build(
                     if (bindingType != null) {
                         soapBinding = bindingType.value().asString();
                     }
-                    cxfWebServices.produce(new CxfWebServiceBuildItem(cxfBuildTimeConfig.path, sei, soapBinding, wsNamespace,
-                            wsName, wrapperClassNames, implementor));
+                    cxfWebServices.produce(new CxfWebServiceBuildItem(
```
shumonsharif, line 259: Is this just a formatting change; if so, kindly revert.

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
```diff
@@ -236,12 +256,23 @@ public void build(
...
+                            soapBinding,
+                            wsNamespace,
+                            wsName,
+                            wrapperClassNames,
+                            implementor));
                 }
-                cxfWebServices.produce(new CxfWebServiceBuildItem(cxfBuildTimeConfig.path, sei, soapBinding, wsNamespace,
-                        wsName, wrapperClassNames));
             }
+            cxfWebServices.produce(new CxfWebServiceBuildItem(
```
shumonsharif, line 269: Kindly revert if this is just a formatting change.

shumonsharif (changes requested):
Hey @wh81752 Ignore my last comment, I think it's just these two changes that are missing ... everything else looks good!

wh81752:
@shumonsharif - This (accidental) formatting changes are killing me. Which IDE are you using? If it happens to be IntelliJ IDEA please be so kind to export format-rules for importing.
Btw, this is my Intellij feature request to solve this matter once and forever: https://intellij-support.jetbrains.com/hc/en-us/requests/3270698    (autodected formatting rules based on current code using AI/machine-learning).

shumonsharif:
Yeah, I use IntelliJ and I never set up any rules; I don't happen to have the issue you keep having  - I think it may possibly be one of your extensions causing it.  BTW, did you just change the formatting in the last commit, or did you modify the actual behavior?  Is there a reason you moved [this](https://github.com/quarkiverse/quarkus-cxf/pull/181/files#diff-ab6082b1f5a32014ec91e75ab14715d38b1d3e2bd2cab287562542c1af6a37b6R263) out of the if / else block?

wh81752:
@shumonsharif - please export the rules so I can import (even if they are plain standard rules).

> Is there a reason you moved this out of the if / else block?

What is the "this"? Don't get it.

shumonsharif:
https://github.com/quarkiverse/quarkus-cxf/blob/f6a66556175859a0ae8d2016f1a965d18eb99249/deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java#L263

shumonsharif:
> @shumonsharif - please export the rules so I can import (even if they are plain standard rules).

I'm happy to export the rules for you, but would that really help?  Both formats (ie. what you had committed and what's in master) are perfectly valid - I believe the issue is not so much the rules ... it's more to do with something in your workspace reformatting everything you touch?  Either way, let's take this up as a separate issue ...

wh81752:
> Is there a reason you moved this out of the if / else block?

Just because it is the same statement, whether it is a client or impl interface (code duplication).

wh81752:
> I'm happy to export the rules for you, but would that really help? Both formats (ie. what you had committed and what's in master) are perfectly valid - I believe the issue is not so much the rules ... it's more to do with something in your workspace reformatting everything you touch? Either way, let's take this up as a separate issue ...

I reverted back to standard IntelliJ rules and I applied them on all Java code. I still see code changed. It would be really good to have some mechanics in place that do the formatting stuff, so that we can discuss the essentials. I really would like to move things forward now.

shumonsharif:
> > Is there a reason you moved this out of the if / else block?
> 
> Just because it is the same statement, whether it is a client or impl interface (code duplication).

Aah, gotcha ... I didn't look close enough, I thought the logic was getting impacted.  I am good with the code as it stands currently ... hopefully @dufoli can do one last review and if things look good, he can merge.  Thanks once again @wh81752 , I think this PR addresses many limitations quite nicely.

shumonsharif (approved):

dufoli:
I will review again but seems goot to merge.
Anyway because of the big change on configuration. I will need to 
- do a new release just before merging to let the time to test master

dufoli:
and please rebase

famod:
@wh81752
> > Something like Instance would come to mind...
> 
> Have not tested this yet. Technically it shoud work, however, I guess you need to figure out to whom you are talking then, i.e. what is the particular config behind an instance?

With `javax.enterprise.inject.Instance<T>` you can call `.select(Annotation... qualifiers)`, so you'd basically pass over qualifier annotation _instances_ (which are a bit awkward to create, but IIRC there is a helper for that somewhere).
Since `@CXFClient` already is a qualifier, I suppose you can simply use instances of that annotation. 👍

floriansiepe:
@wh81752 Great PR! From my point of view (custom) named client are crucial, when having multiple SOAP clients in a project. Since the current configuration assumes a a certain url scheme, which appends the config key at the specified url. E.g.
```
quarkus.cxf.endpoint."/tracking".client-endpoint-url=https://myserviceUrl
quarkus.cxf.endpoint."/tracking".service-interface=com.acme.Tracking
```

would result in `https://myserviceUrl/tracking`.

Unfortunately, this does not working in scenarios, where the multiple services are registered at the "root" path of the host. E.g. 
```
quarkus.cxf.endpoint."/".client-endpoint-url=https://myFirstServiceUrl
quarkus.cxf.endpoint."/".service-interface=com.acme.FirstInterface

quarkus.cxf.endpoint."/".client-endpoint-url=https://mySecondServiceUrl
quarkus.cxf.endpoint."/".service-interface=com.acme.SecondInterface
```

How to distinguish the configuration for these clients? I would be great to see this in the next release!

wh81752:
@famod and @floriansiepe - glad to hear that this PR _appears_ to be useful. Would you mind to read the client docu of this PR (https://github.com/wh81752/quarkus-cxf/blob/pr/client-config-logic/docs/modules/ROOT/pages/client.adoc) as well? I'm not yet that happy with it. Feel free to share any thoughts.

floriansiepe:
> @famod and @floriansiepe - glad to hear that this PR _appears_ to be useful. Would you mind to read the client docu of this PR (https://github.com/wh81752/quarkus-cxf/blob/pr/client-config-logic/docs/modules/ROOT/pages/client.adoc) as well? I'm not yet that happy with it. Feel free to share any thoughts. 

@wh81752 Seems good to me. The only thing which I think could be confusing is the example of fallback values when under the specific name a key does not exists. My thought would be that named cxf properties (just like jdbc datasources) are completely separate units.

wh81752:
@floriansiepe - ahhh, good point:
- cxf (client) properties are completely seperate units, i.e. no unit inherits from another unit or other magic.
- If you use `@CXFClient("foobar")` then  unit `quarkus.cxf.client."foobar"` must exist, otherwise error.

What if you use `@CXFClient`? Then there is indeed a fallback: 
- If there is a config unit matching the SEI then that unit is applied, otherwise 
- some default values (based on `@WebService` properties) are used.
And what if there are more then one config unit matching a SEI? Then a unit marked as "fallback" will be choosen and error if two units are marked as "fallback" per SEI. 

Hmm, looks like I need to reformulate that docu section OR (worse) think again about the design. If something is difficult to explain then its most likely badly designed (sigh).

famod:
@wh81752 Hm, I also read that bit with the _default configuration_ and now I'm not sure anymore whether what I wrote about `Instance` (https://github.com/quarkiverse/quarkus-cxf/pull/181#issuecomment-831553334) will work here.
I'll try to conduct a test with you branch over the weekend, this use case is important for me.

As for the design: I generally like conventions over configuration but maybe something like that would make it more obvious?
```
quarkus.cxf.client."my-featured-fruitservice-client".inherits=my-fruitservice-client
```
But this can become tricky to validate.

Btw, are global defaults possible? E.g.:
```
quarkus.cxf.client.defaults.in-interceptors=...
```
or:

```
quarkus.cxf.client.in-interceptors=...
```

Options are endless... 😁

dufoli:
Hello, mono client as a default producer and a named producer... You can see how it worked but it is great to have default config when not named client is injected

dufoli:
But it can be handled after release... So maybe done later

dufoli:
Can you do git rebase master - i
Then squash all commit by adding s instead of p for all line except the first one.
To avoid all merge issue you can start to squash first by doing git rebase - i head~23 and squash

wh81752:
@dufoli - I must admit that I'm totally lost in rebasing. I'm solving conflicts again which I have solved hours ago.

wh81752:
I propose to merge into master now and doing a rebase on the master itself. Life is too short to wrestle around with git defects.

dufoli:
OK I will do it

dufoli:
I have release and review to add a lost feature in merge.
I merge the PR just after the release process. @shumonsharif  you can do it either. Because I will start my real job. So maybe do the merge this evening. But wait for the release.
Many thanks @wh81752 for your hard work and time to produce this huge patch.

dufoli:
FYI, rebase was too compexe because merge has been done previously. I have think to it for next time we have a such patch, it has take time and I think I take the wrong approach.
- on feature branch next time continue your work without merge/rebase
- at the end rebase -i on itself to squash (or autosquash)
- and when a single commit is there, rebase on master so we do not lost history and all intermediate commit is working.

 squash to avoid to have too much commit and rebase/merge at the end. Here we have a spagetti because

wh81752:
> FYI, rebase was too compexe because merge has been done previously. I have think to it for next time we have a such patch, it has take time and I think I take the wrong approach.
> 
> on feature branch next time continue your work without merge/rebase
> at the end rebase -i on itself to squash (or autosquash)
> and when a single commit is there, rebase on master so we do not lost history and all intermediate commit is working.
> squash to avoid to have too much commit and rebase/merge at the end. Here we have a spagetti because

@dufoli - thx for your hard work on merging this PR. I must admit that I got lost in tracking incoming changes.

I really wonder what the best strategy is or rather would have been. Not "getting" changes from master appears also bit strange. 

I guess that instead of "merging in" master changes I should have "rebased" my branch on master's HEAD. This way my feature branch consists of feature branch commits only.

And, as you mentioned, squash feature branch commits in large units (ideally just one). I guess the main reason is that when you have a feature with N commits and each commit changes a file also changed in master, then you end up in N merge conflicts, i.e. you resolve over and over again until you start to freak out (like I did :-)
 
Once again thanks for merging.

PR #188, opened 2021-04-07
Pr/named client

wh81752:
Dear All,

this is an alternative implemention of #181 . Instead of @CXFClient standard qualifier @Named is used to select a particular client configuration:

```
import javax.inject.Inject;
import javax.inject.Named;
import com.example.FruitWebService;

@Inject
@Named("my-featured-fruitservice-client")
FruitWebService clientService;   // ok, inject featured client
```

It also contains updated client documentation as well as many unit and integration tests.


wh81752:
I'm getting more and more convinced that using `@Named` is not a good idea and rather introduce dedicated annotations `CXFClient` and` CXFImpl.`

One mayor problem is this one:

```
@Inject @Named("foo") FruitWebService foo;        // A
@Inject @Named("bar") FruitWebService bar;        // B

@Inject @Default FruitWebService foobar;             // C
```

The injection at point (C) is ambigious and appears unresolvable using qualifier `@Named`. Usually qualifier `@Default` gets added to a producer method _if no other qualifier_ is present. Hence this would allow for the creation of a "default" producer method applicable at injection point (C) to resolve that ambigiouty (by selecting a default configuration for that service).

Unfortunatly, `@Named` is an exception to the rule above: `@Default` gets added even in the presence of qualifier `@Named`. Hence the code above reads:

```
@Inject @Default  @Named("foo") FruitWebService foo;        // A
@Inject @Default  @Named("bar") FruitWebService bar;        // B

@Inject @Default FruitWebService foobar;                               // C
```

and this also applies for producer methods, here producer "foo":

```
@Produces @Named @Default                // CDI automatically added @Default (!!)
FruitWebService foo() { return .. }
```

Now it's obvious that producer `foo()` can be applied to (C).

Another problem with `@Named` is that it's annotation value is **bound**. Hence if we have `@Named("foo")` then a bean or producer named `foo` must exist. To overcome this, every injection point must be investigated, filtered and appropriate producers must be created (=> this is partially done in this PR where only field injection points are handled).

The problem with this approach is this one - assume for the moment that we have (in a very large and complex webapp) only a named "foo"injection point:

```
@Inject @Named("foo") FruitWebService foo;        // A
@Inject FruitWebService foobar;                              // C
```

Now there is only one producer and created and hence injection at (C) works without problem. Well, one problem is that you get configuration "foo" injected at (C) without further indication. Assume that someone changes (A)  to use `@Named("bar")` - now you also get "bar" injected at (C). This is rather non-intuitive and I guess very difficult to discover.

Then, again some time later, another team member introduces `@Named("foo")` again at a difference code location creating an ambigiouty at (C) - hence the application does not even startup now.

All this makes me believe that `@Named` is an inappropriate choice.

Better to have something like `@CXFClient` and `@CXFImpl` as qualifiers. This names tell exactly what is injected, making the code readable. The annotation property of  `@CXFClient` is also unbound (see #181) meaning that only one producer method is required and no static injection point scanning is necessary. This simplfies coding and especially testing.

It also avoids the problem described above:

```
@Inject @CXFClient("foo") FruitWebService foo;                  // A
@Inject @CXFClient            FruitWebService foobar;             // C
..
@Inject @CXFClient("bar") FruitWebService bar;                  //B
```

Here always the very same configured client is injected at (C) regardless of whether (B) is added later. What is injected at (C) is the default client where the default is a pure configuration issue. There is also no ambigiouty introduced by (B) later on cause there is only one producer method anyway.

wh81752:
I wonder how to continue with this PR and PR #181 ?

dufoli:
we will use the other PR... so close this one @shumonsharif  ok ?

PR #210, opened 2021-05-03
merge resource for uberjar and native

### LINKED ISSUE #194: Packaged application gives error message: Could not resolve a binding for http://schemas.xmlsoap.org/wsdl/soap/
When using the application with dev mode ( without packaging ) everything works perfectly. But after we package it as an uber-jar it gives the following error:
````
Could not resolve a binding for http://schemas.xmlsoap.org/wsdl/soap/
````

We are using quarkus-cxf version 0.5 and quarkus version 1.11.3.Final.

If you need more information, please let me know.

dufoli:
fix #194


dufoli:
solution do not worked so cancel this PR.

PR #214, opened 2021-05-10
Dependabot: ignore jakarta.xml.ws-api >= 3.0.0

famod:
Should prevent PRs like #200

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,7 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    ignore:
+      - dependency-name: jakarta.xml.ws:jakarta.xml.ws-api
+        versions:
+          - ">= 3.0.0"


shumonsharif:
@famod @dufoli I think our jakarta.xml.ws-api dependency version should depend on what the CXF libraries depend on.  I'm not sure if CXF plans to update to anything >= 3.0.0 but in case that happens in the future, won't this PR impact us at that point?

shumonsharif:
Gotcha, makes sense @famod.  In a way, it is indeed better that this dependency is not managed by dependabot as it can possibly lead to incongruent versions.  We'll just need to keep it in mind when we update CXF.  Including this link here for reference:  https://cxf.apache.org/docs/migration-guides.html

famod:
I actually went for `>= 0` in a Maven extension I'm maintaining: https://github.com/gitflow-incremental-builder/gitflow-incremental-builder/blob/master/.github/dependabot.yml
It just doesn't make sense to let the bot (try to) bump everything.

shumonsharif:
Yes, I would suggest we do the same here as well.  @dufoli wdyt?

PR #226, opened 2021-05-20
cleanup dependency and remove asm handled by gizmo

### LINKED ISSUE #224: ASM dependency?
(Maybe GH discussions should be enabled for this repo, which is now possible as an opt-in for quarkiverse projects.)

I noticed that this extension brings in `org.ow2.asm:asm:jar:9.1:compile`. But I'd assume that all bytecode magic is done during build time instead of runtime?

dufoli:
fix #224

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,6 +28,7 @@
         <developer>
             <id>dufoli</id>
             <name>Olivier Dufour</name>
+            <email>olivier dot duff at gmail(DOT)com</email>
         </developer>
         <developer>
             <id>dufgui</id>
@@ -57,6 +58,7 @@
         <maven-compiler-plugin.version>3.8.1</maven-compiler-plugin.version>
         <maven-surefire-plugin.version>2.22.2</maven-surefire-plugin.version>
         <maven-failsafe-plugin.version>2.22.2</maven-failsafe-plugin.version>
+        <jaxb-runtime.version>2.3.3-b02</jaxb-runtime.version>
     </properties>
 
     <scm>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -109,7 +115,18 @@
                 <groupId>jakarta.xml.ws</groupId>
                 <artifactId>jakarta.xml.ws-api</artifactId>
                 <version>${jakarta.xml.ws-api.version}</version>
-            </dependency>     
+            </dependency>
+            <dependency>
+                <groupId>org.glassfish.jaxb</groupId>
+                <artifactId>jaxb-runtime</artifactId>
```
famod, line 121: There is also `quarkus-jaxb`... 🤔

shumonsharif: Good point @famod we already have a dependency on quarkus-jaxb, which has everything we need; I have done some additional cleanup.

shumonsharif:
@dufoli @famod Please review latest updates.

dufoli:
@shumonsharif  Maybe add quarkus-jaxb to main pom in dependencyManagement to have all external dep in main pom wdyt ?. else it is ok for me.

shumonsharif:
Hi thanks @dufoli the `quarkus-bom` dependency in the main pom in dependencyManagement is what makes `quarkus-jaxb` available.  Even if we define in dependencyManagement, we'll still need it also in the `runtime/pom.xml` as well.

famod (approved):
LGTM!

famod:
Btw, if you want to make extra sure this doesn't slip in again, then you could add something like https://maven.apache.org/enforcer/enforcer-rules/bannedDependencies.html

I would also suggest https://maven.apache.org/enforcer/enforcer-rules/dependencyConvergence.html but that addresses another category of problems.

PR #232, opened 2021-05-22
doc release branch in github action

dufoli:


dufoli:
@gastaldi  @shumonsharif  I am not expert in github action. Is it the right way to create branch v0.7 v0.8... for doc release

shumonsharif:
@dufoli I think this is along the right lines, and should work good ... but there are a couple of steps missing.

The quarkiverse-docs build will fail if you don't update the following line to reflect the correct version, and the docs won't get generated.
https://github.com/quarkiverse/quarkus-cxf/blob/v0.7/docs/antora.yml#L3

Something like this should do the trick:
https://github.com/marketplace/actions/substitute-string#examples

Once we update the version in antora.yml, we'll need to commit the change to the new branch before we push.

shumonsharif:
Actually, this will be a better option for a regex replace, and it should be limited to antora.yml.
https://github.com/marketplace/actions/find-and-replace

shumonsharif:
I'm not 100% if this will work, but I think the following may be the cleanest solution, without any additional actions:

```yaml
      - name: doc branch release ${{steps.metadata.outputs.current-version}}
        run: |
          git checkout -b v${{steps.metadata.outputs.current-version}}
          sed -i -E "s/version:.*/version: ${{steps.metadata.outputs.current-version}}/g" docs/antora.yml
          git commit -m "Updated version in antora.yml" docs/antora.yml

      - name: Push branch
        uses: ad-m/github-push-action@v0.6.0
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
          branch: ${{github.ref}}
```

gastaldi:
How about using a badge or some kind of marker to identify that a certain feature works on a specific version onwards? That would eliminate the need to version the docs and introduce separate branches AFAIK?

shumonsharif:
@gastaldi The issue is not so much a new feature getting released that we can easily mark in the docs ... it's more that the new release will introduce breaking, non backward compatible changes which will make the previous docs completely invalid.  The docs for the two releases won't coexist very well in one doc.

We don't plan on maintaining the older branches, we'll periodically clean them up, if that helps.

dufoli:
@shumonsharif indeed the situation is only between 0.7 and master because we have configuration change. So lets cancel this PR. and create branch only when we need it.

PR #247, opened 2021-06-10
Ban bind-api and asm from runtime module

famod:
Those are excluded since a while in root pom but are required for the deployment module (coming from other dependencies).

I removed the exclusion for `javax.xml.soap-api` since that didn't do anything, `cxf-rt-frontend-jaxws` does not depend on this (anymore?).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,10 +88,6 @@
                 <version>${cxf.version}</version>
                 <exclusions>
                     <exclusion>
-                        <groupId>javax.xml.soap</groupId>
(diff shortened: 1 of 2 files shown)


famod:
I can ban `javax.xml.soap-api` as well if you like, to make sure it doesn't pop up again in the future.

shumonsharif:
@famod I think your PR looks good, and is better as-is.  Banning `javax.xml.soap-api` without an exclusion may lead to confusion down the road.

shumonsharif (approved):

PR #260, opened 2021-06-30
Drop Java 8, move on to 11

famod:
Quarkus 2 dropped Java 8 and moved on to Java 11, so there is no use sticking to 8 for this extension.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,8 +47,9 @@
 
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+        <maven.compiler.release>11</maven.compiler.release>
         <maven.compiler.parameters>true</maven.compiler.parameters>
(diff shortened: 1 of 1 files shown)


famod:
@gastaldi are we expecting all quarkiverse maintainers to realize that on their own? (no offense, just asking)

gastaldi:
> @gastaldi are we expecting all quarkiverse maintainers to realize that on their own? (no offense, just asking)

@famod no offense taken, I think we could specify that parameter in the quarkiverse-parent, WDYT?

famod:
> I think we could specify that parameter in the quarkiverse-parent, WDYT?

@gastaldi yes, makes sense. My primary concern were existing extensions, but I guess we can expect maintainers to read Quarkus blog posts and changelogs carefully?

famod:
Stupid me...in the parent...which is picked up by dependabot (no?). That would cover existing extensions as well but maintainers would need to drop their properties.
So I guess it's mainly a question about how to communicate this. 🙂

gastaldi:
> Stupid me...in the parent...which is picked up by dependabot (no?). 

That is correct.

> That would cover existing extensions as well but maintainers would need to drop their properties.
> So I guess it's mainly a question about how to communicate this.


Right, I think sending a message to quarkus-dev is the preferred choice (assuming the extension maintainers are subscribed to that - which I believe they are :))

gastaldi:
I think we should also generate a Changelog in the quarkiverse-parent, so that is automatically picked by Dependabot

famod:
That's a good idea! Self-explaining update FTW!

PR #261, opened 2021-07-01
add a test and handle http rootPath but do not worked...

### LINKED ISSUE #124: cannot get to wsdl when application.properties has quarkus.http.root-path
The moment I add quarkus.http.root-path (which I need for my rest endpoints) to application.properties, I cannot get to the wsdl anymore. 

Replicate: Use the example and add quarkus.http.root-path=/test

dufoli:
fix #124

### DIFF
--- a/runtime/src/main/java/io/quarkiverse/cxf/transport/CxfHandler.java
+++ b/runtime/src/main/java/io/quarkiverse/cxf/transport/CxfHandler.java
@@ -158,22 +158,6 @@ public class CxfHandler implements Handler<RoutingContext> {
         }
     }
 
-    private String mergePath(String servletPath, String relativePath) {
-        if (servletPath.endsWith("/")) {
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/io/quarkiverse/cxf/transport/CxfHandler.java
```diff
@@ -156,6 +158,22 @@ public CxfHandler(CXFServletInfos cxfServletInfos, BeanContainer beanContainer)
...
+            }
+        } else if (!relativePath.equals("/")) {
+            if (!relativePath.startsWith("/")) {
+                servletPath += "/";
+            }
+            servletPath += relativePath;
+        }
+        return servletPath;
+    }
+
```
dufoli, line 176:
```suggestion


```

PR #262, opened 2021-07-01
Limit client generation to only those portTypes that are actually used

### LINKED ISSUE #252: Client: Should limit generation to only those portTypes that are actually used
We generate classes via `cxf-codegen-plugin` from a WSDL file that contains multiple portTypes.
This yields multiple `@WebService` classes.
So far so common.

We only use one of those service interfaces/portTypes but I can see on the console that this extension generates a  QuarkusJaxWsServiceFactoryBean for _every_ found `@WebService` class.

Since Quarkus is all about build-time optimization and generation is expensive, I think it makes a lot of sense to limit this generation to only those portTypes that are actually in use.

famod:
Resolves #252

Includes some cleanups and refactorings to QuarkusCxfProcessor.

Also introduces tests that check constructor and instance injection.

A general note: I tried to reduce the number of local variables in the processor to make the code more readable (IMO).

### DIFF
--- a/deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
+++ b/deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
@@ -11,18 +11,20 @@ import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
 import java.lang.reflect.Modifier;
 import java.net.URL;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.inject.Instance;
 import javax.enterprise.inject.Produces;
 import javax.enterprise.inject.spi.InjectionPoint;
 import javax.inject.Inject;
@@ -52,6 +54,7 @@ import org.jboss.jandex.ClassInfo;
 import org.jboss.jandex.DotName;
 import org.jboss.jandex.IndexView;
 import org.jboss.jandex.MethodInfo;
+import org.jboss.jandex.MethodParameterInfo;
 import org.jboss.jandex.Type;
 import org.jboss.logging.Logger;
 import org.w3c.dom.Document;
@@ -111,6 +114,8 @@ import io.vertx.ext.web.RoutingContext;
 class QuarkusCxfProcessor {
 
     private static final String FEATURE_CXF = "cxf";
+    private static final DotName CXFCLIENT_ANNOTATION = DotName.createSimple(CXFClient.class.getName());
+    private static final DotName INJECT_INSTANCE = DotName.createSimple(Instance.class.getName());
     private static final DotName WEBSERVICE_ANNOTATION = DotName.createSimple("javax.jws.WebService");
     private static final DotName WEBSERVICE_CLIENT = DotName.createSimple("javax.xml.ws.WebServiceClient");
     private static final DotName REQUEST_WRAPPER_ANNOTATION = DotName.createSimple("javax.xml.ws.RequestWrapper");
@@ -138,7 +143,12 @@ class QuarkusCxfProcessor {
         }
     }
 
-    private String getNamespaceFromPackage(String pkg) {
+    private String getNameSpaceFromClassInfo(ClassInfo wsClassInfo) {
+        String pkg = wsClassInfo.name().toString();
+        int idx = pkg.lastIndexOf('.');
+        if (idx != -1 && idx < pkg.length() - 1) {
+            pkg = pkg.substring(0, idx);
+        }
         //TODO XRootElement then XmlSchema then derived of package
         String[] strs = pkg.split("\\.");
         StringBuilder b = new StringBuilder("http://");
@@ -212,7 +222,7 @@ class QuarkusCxfProcessor {
             BuildProducer<GeneratedBeanBuildItem> generatedBeans,
             BuildProducer<CxfWebServiceBuildItem> cxfWebServices,
             BuildProducer<AdditionalBeanBuildItem> additionalBeans,
-            BuildProducer<UnremovableBeanBuildItem> unremovableBeans) {
+            BuildProducer<UnremovableBeanBuildItem> unremovableBeans) throws ClassNotFoundException {
         IndexView index = combinedIndexBuildItem.getIndex();
         // Register package-infos for reflection
         for (AnnotationInstance xmlNamespaceInstance : index.getAnnotations(XML_NAMESPACE)) {
@@ -220,9 +230,12 @@ class QuarkusCxfProcessor {
                     new ReflectiveClassBuildItem(true, true, xmlNamespaceInstance.target().asClass().name().toString()));
         }
 
-        //TODO bad code it is set in loop but use outside...
-        ClassOutput classOutput = new GeneratedBeanGizmoAdaptor(generatedBeans);
-        quarkusCapture c = new quarkusCapture(classOutput);
+        Bus bus = BusFactory.getDefaultBus();
+        // setup class capturing
+        bus.setExtension(new quarkusCapture(new GeneratedBeanGizmoAdaptor(generatedBeans)),
+                GeneratedClassClassLoaderCapture.class);
+
+        Set<String> clientSEIsInUse = findClientSEIsInUse(index);
 
         for (AnnotationInstance annotation : index.getAnnotations(WEBSERVICE_ANNOTATION)) {
             if (annotation.target().kind() != AnnotationTarget.Kind.CLASS) {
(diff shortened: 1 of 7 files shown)


deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
```diff
@@ -212,93 +222,94 @@ public void build(
             BuildProducer<GeneratedBeanBuildItem> generatedBeans,
             BuildProducer<CxfWebServiceBuildItem> cxfWebServices,
             BuildProducer<AdditionalBeanBuildItem> additionalBeans,
-            BuildProducer<UnremovableBeanBuildItem> unremovableBeans) {
+            BuildProducer<UnremovableBeanBuildItem> unremovableBeans) throws ClassNotFoundException {
```
famod, line 225: See discussion #259

famod:
I forgot to push a new test, will do ASAP.

deployment/src/test/java/io/quarkiverse/cxf/deployment/test/CxfClientConstructorInjectionBean.java
```diff
@@ -0,0 +1,31 @@
...
+
+@ApplicationScoped
+public class CxfClientConstructorInjectionBean {
+
+    private final CXFClientInfo clientInfo;
+    private final FruitWebService clientProxy;
+
+    @Inject
+    public CxfClientConstructorInjectionBean(
+            // @Named is omitted her because not required
```
famod, line 17: typo 🤦, will adjust if other things pop up during review

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
```diff
@@ -138,7 +143,12 @@ public void generateWSDL(BuildProducer<NativeImageResourceBuildItem> ressources,
         }
     }
 
-    private String getNamespaceFromPackage(String pkg) {
+    private String getNamespaceFromPackage(ClassInfo wsClassInfo) {
```
dufoli, line 146: can you change it to getNameSpaceFromClassInfo

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
```diff
@@ -212,93 +222,94 @@ public void build(
...
-            String wsName = "";
-            if (annotation.value("serviceName") != null) {
-                wsName = annotation.value("serviceName").asString();
-            }
-            Collection<ClassInfo> implementors = index.getAllKnownImplementors(DotName.createSimple(sei));
+
+            // created on-demand if an implementor or client usage is found
+            QuarkusJaxWsServiceFactoryBean factoryBean = null;
+
+            String wsNamespaceDefault = Optional.ofNullable(annotation.value("targetNamespace"))
```
dufoli, line 261: why call it default if it is always the one we use ? wsNamespace seems fine.

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
```diff
@@ -212,93 +222,94 @@ public void build(
...
+                    .map(AnnotationValue::asString)
+                    .orElseGet(() -> getNamespaceFromPackage(wsClassInfo));
+            String wsNameDefault = Optional.ofNullable(annotation.value("serviceName"))
```
dufoli, line 264: same for wsNameDefault we can keep wsName to avoid new string for nothing.

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
```diff
@@ -138,7 +143,12 @@ public void generateWSDL(BuildProducer<NativeImageResourceBuildItem> ressources,
         }
     }
 
-    private String getNamespaceFromPackage(String pkg) {
+    private String getNamespaceFromPackage(ClassInfo wsClassInfo) {
+        String pkg = wsClassInfo.name().toString();
+        int idx = pkg.lastIndexOf('.');
+        if (idx != -1 && idx < pkg.length() - 1) {
+            pkg = pkg.substring(0, idx);
+        }
```
dufoli, line 151: yes far better to do it like that ;-)

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
```diff
@@ -212,93 +222,94 @@ public void build(
...
+                factoryBean = factoryBean == null ? createQuarkusJaxWsServiceFactoryBean(sei, bus) : factoryBean;
+
+                AnnotationInstance webserviceClient = findWebServiceClientAnnotation(index, wsClassInfo.name());
+                String wsNamespace = wsNamespaceDefault;
+                String wsName = wsNameDefault;
+                if (webserviceClient != null) {
+                    wsName = webserviceClient.value("name").asString();
+                    wsNamespace = webserviceClient.value("targetNamespace").asString();
+                }
+                cxfWebServices.produce(new CxfWebServiceBuildItem(cxfBuildTimeConfig.path, sei, soapBindingDefault, wsNamespace,
```
dufoli, line 303: it is soapBinding and not soapBindingDefault I think no ?

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
```diff
@@ -345,6 +376,16 @@ private AnnotationInstance findWebServiceClientAnnotation(IndexView index, DotNa
...
+    private QuarkusJaxWsServiceFactoryBean createQuarkusJaxWsServiceFactoryBean(String sei, Bus bus)
+            throws ClassNotFoundException {
+        QuarkusJaxWsServiceFactoryBean jaxwsFac = new QuarkusJaxWsServiceFactoryBean();
+        jaxwsFac.setBus(bus);
+        //TODO here add all class
+        jaxwsFac.setServiceClass(Thread.currentThread().getContextClassLoader().loadClass(sei));
+        jaxwsFac.create();
+        return jaxwsFac;
+    }
+
```
dufoli, line 388: I like that ;-)

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
dufoli: hmm ok for me

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
famod: Sure, good point! I wanted to change the name anyway but forgot to do so.

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
famod: I did this because I had a version where overwrite effects could happen by multiple implementors. But it doesn't seem to be a problem anymore.
I'll re-check.

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
famod: Same reason as for `wsNamespaceDefault`. I'll re-check this again.

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
famod: Each implementor can set a another `soapBinding`, that's why I introduced `soapBindingDefault` to avoid that one implementor influences another implementor or a client.

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
famod: I think this must stay: `wsName` is calculated _per implementor_ (= per `wsClass`) and if that overwrites a single `wsName` that is set per outer loop (per `sei`), then it can leak into a client (when it doesn't use `@WebServiceClient`).

It's probably hard to see just here on GH. Better check it out locally.

But I just realized that `wsNameDefault` is never used for implementors; each implementor will always gate it's own `wsName`.
I'll split this up.

famod:
I pushed a sparate commit to make reviewing easier. Will squash it once review is done.

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
famod: done

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
famod: done

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
famod: done

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
famod: PS: the is no adjustment for the client, that's why it uses `soapBindingDefault` directly

dufoli: Strange that cxf clients can not been on soap 1.2 maybe we need configuration for that part.  Or detect this annotation on sei and not only implementor

famod: Haven't you introduced a SOAP 1.2 config option for clients not so long ago? #209

famod:
@shumonsharif can you review as well, please? Thanks!

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
dufoli: Lol I forget about it.... I am dory!

dufoli:
It is OK for me.

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
famod: dory 😄 

But I must say I'm a bit confused what that soap handling TODO in QuarkusCxfProcessor is about then.
Is it about a _global_ default that should be configurable?

shumonsharif:
> @shumonsharif can you review as well, please? Thanks!

@famod @dufoli Apologies for being MIA - I am reviewing now!

shumonsharif (commented):
@famod Looks good to me, but just had a couple questions ... and readability is certainly massively improved with everything you touched!  Thanks so much.

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
```diff
@@ -212,93 +222,89 @@ public void build(
...
             //TODO add soap1.2 in config file
-            String soapBinding = SOAPBinding.SOAP11HTTP_BINDING;
-            //if no implementor, it mean it is client
-            if (implementors == null || implementors.isEmpty()) {
+            final String soapBindingDefault = SOAPBinding.SOAP11HTTP_BINDING;
+
+            Collection<ClassInfo> implementors = index.getAllKnownImplementors(DotName.createSimple(sei));
+
+            if (implementors != null && !implementors.isEmpty()) {
+                factoryBean = factoryBean == null ? createQuarkusJaxWsServiceFactoryBean(sei, bus) : factoryBean;
```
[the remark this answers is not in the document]
famod: The creation of the `factoryBean` is lazy to make sure it is only created when _really required_, that is when an implementor and/or a client is found.

famod:
I have just squashed the two commits.

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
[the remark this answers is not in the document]
famod: See also these two lines further up: https://github.com/quarkiverse/quarkus-cxf/pull/262/files#diff-ab6082b1f5a32014ec91e75ab14715d38b1d3e2bd2cab287562542c1af6a37b6R258-R259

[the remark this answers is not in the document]
shumonsharif: What I meant is that factoryBean will always be null on line 271 (so it's a redundant check) - and I believe for line 286 possibly as well?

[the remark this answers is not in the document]
famod: No, because there is a loop over (possibly) multiple implementors and the factoryBean shall be:
- created laziliy
- and only once

So I have to cover:
- neither impl nor client -> don't create bean at all
- create once on first impl and reuse for subsequent impls (and client, if required)
- create for client in case no impl is found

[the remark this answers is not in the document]
shumonsharif: OK, I may be seeing something wrong then, my apologies if so ... what I'm seeing is that you are setting the factoryBean to null (line 259) **inside the for loop, and for every iteration** - so I don't know how you are for example, creating it on first impl and reusing for subsequent impls ...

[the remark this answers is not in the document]
famod: Watch out, there are two nested loops! (which might be another candidate for refactoring)

You have an outer loop for each SEI and per SEI there is also a loop over all impls of that SEI.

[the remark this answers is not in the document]
shumonsharif: I think as long as you're sure of the flow here, we can resolve this ... I just wasn't seeing the case where factoryBean would be something other than null!  But you've verified, so I think we're good!

shumonsharif (approved):

dufoli:
LGTM ? 
@shumonsharif  if you are ok you can merge ;-)

dufoli:
and even release ;-)

shumonsharif:
Sounds good @dufoli ... I'm on it!

PR #315, opened 2021-11-13
X-Forwarded headers implementation

### LINKED ISSUE #311: Managing x-forwarded headers
Hello, 

First of all, thanks for your work, this extension is really making the job.

Is there a way to manage x-forwarded-headers? If not, can I have advice implementing this feature? 

CXF servlet manage them using a flag, authorizing application to be deployed behind an API gateway, and looks like Quarkus has already the capability to do this (https://quarkus.io/guides/http-reference#reverse-proxy).

Regards, 

Guillaume Demeure

shumonsharif:
Fixes #311

### DIFF
--- a/integration-tests/src/main/resources/junit-platform.properties
+++ b/integration-tests/src/main/resources/junit-platform.properties
@@ -1,0 +1,1 @@
+junit.jupiter.testclass.order.default=org.junit.jupiter.api.ClassOrderer$ClassName
--- a/integration-tests/src/main/resources/junit-platform.properties
+++ b/integration-tests/src/main/resources/junit-platform.properties
@@ -1,1 +1,1 @@
-junit.jupiter.testclass.order.default=org.junit.jupiter.api.ClassOrderer$ClassName
+junit.jupiter.testclass.order.default=io.quarkus.test.junit.util.QuarkusTestProfileAwareClassOrderer
--- a/runtime/src/main/java/io/quarkiverse/cxf/transport/CxfHandler.java
+++ b/runtime/src/main/java/io/quarkiverse/cxf/transport/CxfHandler.java
@@ -263,6 +263,13 @@ public class CxfHandler implements Handler<RoutingContext> {
         return reqPrefix;
     }
 
+    /**
+     * Leverages the Quarkus HTTP proxy configuration properties,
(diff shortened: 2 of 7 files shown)


shumonsharif:
@famod Would you be kind enough to help with a quick code review for this PR?  For some odd reason, I couldn't see your GitHub handle to add you in the 'Reviewers' section.

integration-tests/src/main/resources/junit-platform.properties
```diff
@@ -0,0 +1 @@
+junit.jupiter.testclass.order.default=org.junit.jupiter.api.ClassOrderer$ClassName
```
famod, line 1: FYI, I added a `@TestProfile` aware orderer to Quarkus in...hum...2.3? It's mentioned here: https://quarkus.io/guides/getting-started-testing#testing_different_profiles

runtime/src/main/java/io/quarkiverse/cxf/transport/CxfHandler.java
```diff
@@ -251,6 +263,32 @@ private String getBaseURL(HttpServerRequest request) {
         return reqPrefix;
     }
 
+    private HttpServletRequest checkXForwardedHeaders(HttpServletRequest request) {
```
famod, line 266: Might make sense to reference https://github.com/apache/cxf/blob/cxf-3.4.5/rt/transports/http/src/main/java/org/apache/cxf/transport/servlet/AbstractHTTPServlet.java#L301

famod (approved):
LGTM

integration-tests/src/main/resources/junit-platform.properties
shumonsharif: Awesome, a much better option, and should help as we add more test profiles.  Thank you!

runtime/src/main/java/io/quarkiverse/cxf/transport/CxfHandler.java
shumonsharif: Excellent point - I added a brief comment as well that says the method is leveraging the [Quarkus HTTP proxy configuration properties](https://quarkus.io/guides/http-reference#reverse-proxy), instead of purely HTTP headers.  Will need to get all added to the docs.

Issue #211, opened 2021-05-04
Add timeout properties


floriansiepe:
# Feature Request

For each endpoint (in client mode) it should be possible to set some properties on the HTTPClientPolicy such as connection and receive timeouts to make this extension even more useful and customizable.

Please also see:

https://cxf.apache.org/docs/client-http-transport-including-ssl-support.html

shumonsharif:
Hi @floriansiepe I haven't tested any of this, but as a workaround, you may be able to do something like the below.

```java
HTTPConduitConfigurer httpConduitConfigurer = new HTTPConduitConfigurer() {
    public void configure(String name, String address, HTTPConduit c) {
        HTTPClientPolicy httpClientPolicy = new HTTPClientPolicy();
             
        httpClientPolicy.setConnectionTimeout(36000);
        httpClientPolicy.setAllowChunking(false);
        httpClientPolicy.setReceiveTimeout(32000);
              
        c.setClient(httpClientPolicy);
    }
}

final Bus bus = BusFactory.getThreadDefaultBus();
bus.setExtension(httpConduitConfigurer, HTTPConduitConfigurer.class);
```

floriansiepe:
Hi @shumonsharif  thank you very much for the code snippet. This just works fine! 
However using properties instead of code would be great

dufoli:
have you try regular http quarkus config ?

quarkus.http.auth.form.timeout
quarkus.http.idle-timeout
quarkus.http.read-timeout

shumonsharif:
> Hi @shumonsharif thank you very much for the code snippet. This just works fine!
> However using properties instead of code would be great

@floriansiepe @dufoli My suggestion would be to keep this extension focused on the core features provided by CXF.  There are potentially hundreds of CXF configurations we can expose through properties ... the extension will become quite difficult to manage if we try to address all of them.

As per the link @floriansiepe shared, this is an advanced configuration use case for CXF that can easily be handled through code.  While managing SLAs in the SOAP world may be a common enough use case, I'm not sure how much value it adds to be configurable through properties - maybe we can leave the issue open to gauge community interest?

dufoli:
Fair point @shumonsharif. So let's close this issue

floriansiepe:
@shumonsharif I see your point. Certainly, making every possible piece of CXF configurable in the extension is not beneficial for anyone. 

[...]

PR #325, opened 2021-12-04
set client connection and receive timeout via property

arolfes:


shumonsharif:
Hi @arolfes - thanks so much for the PR.  We've had this topic of config properties for connect and receive timeouts [come up in the past](https://github.com/quarkiverse/quarkus-cxf/issues/211).  We thought it best to not include it as part of this extension, due to reasons including:

- There are 25 or so [properties](https://cxf.apache.org/docs/client-http-transport-including-ssl-support.html#ClientHTTPTransport(includingSSLsupport)-Theclientelement) that can be configured in `HTTPClientPolicy`.  Accounting for only two of them is not ideal, and sets a precedent for the rest of them to also be configurable through the extension properties.
- Similarly, there are numerous [configurations](https://cxf.apache.org/docs/client-http-transport-including-ssl-support.html#ClientHTTPTransport(includingSSLsupport)-Theconduitelement) for `HTTConduit`.   Merging and/or overriding custom application specific HTTPConduit configurations with what we set via this extension can become challenging.
- Given the points above, and as we have simple and [documented](https://quarkiverse.github.io/quarkiverse-docs/quarkus-cxf/dev/client.html#code-config) ways to set the `HTTConduit` and `HTTPClientPolicy` on the CXF bus - we thought it best to leave connection related advanced configurations to the application developers.

I am still of the sentiment that we should not include these properties as part of the extension, but maybe let's keep this open for now to gauge community interest and get any other community feedback.

arolfes:
Hi @shumonsharif ,

I understand your points and I agree with it. 
I close this PR. 

Regards

PR #338, opened 2021-12-28
XXE Mitigation & Code Cleanup

shumonsharif:


shumonsharif:
@famod Would you be kind enough to do a quick code review here. 

Nothing too drastic in terms of code cleanup, BUT there are a couple XXE related security issues that are addressed with how we were using `DocumentBuilderFactory` and `TransformerFactory` in `QuarkusCxfProcessor`.  Both issues were identified in a local Sonar scan, more details at the link below:

https://rules.sonarsource.com/java/RSPEC-2755

famod:
Can you comment on the removed classes? IIRC, at some point I had a look at `QuarkusJaxWsProxyFactoryBean` (don't remember why) so now I'm surprised it's gone.

shumonsharif:
> Can you comment on the removed classes? IIRC, at some point I had a look at `QuarkusJaxWsProxyFactoryBean` (don't remember why) so now I'm surprised it's gone.

Thanks @famod I used the jacoco agent and jacoco cli libraries to put together some coverage reports (for both unit tests and, more importantly, runtime integration tests).  These three deleted classes (`VertxConduit`, `QuarkusJAXBBeanInfo` and `QuarkusJaxWsProxyFactoryBean`) were not used at all in any of the flows.  

Also, searching through references in the codebase confirms the same.  For example, if we search for `QuarkusJaxWsProxyFactoryBean` in the [existing code](https://github.com/quarkiverse/quarkus-cxf/search?q=QuarkusJaxWsProxyFactoryBean), there are no Java references to the class, and nothing indicating that it's used in the native flows either.  So, I concluded that these classes may have been used at some point in the past, but they are currently no longer needed.  Let me know your thoughts!

famod (approved):
LGTM! Only found a small leftover:
https://github.com/quarkiverse/quarkus-cxf/blob/5e832f47a11f4b0812d767995caf5411e9d82221/deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java#L756
This should be removed as well.

shumonsharif:
Awesome, thank you - I had removed the leftover bit in a subsequent [commit](https://github.com/quarkiverse/quarkus-cxf/pull/338/commits/5b1532974e119a060ed4b5e83f555773f40e85e7).  Thanks again!

dufoli:
@shumonsharif I just read all commit since I left and good catch. BTW to explain the history, I create VertxConduit because of cxf doc about how to build a custom VertxConduit because when I need to plus vertx to cxf without other http server, I used this doc : https://cxf.apache.org/docs/custom-transport.html
and it need to build conduit and transport but conduit is for client side so I do not need a custom one to switch to vertx and I forget to clean it up.

shumonsharif:
@dufoli It's great to hear from you!  Are you back?  Or just dropping in for a quick visit?  :)  Hope all is well.

Issue #340, opened 2022-01-08
WSS4J dependencies


EmreKp:
Hello,

In order to use `wss4j`, should we have only add a dependency like `cxf-rt-ws-security`? Is there any other way to use it, or is this the easiest way? If this is only way, documentation also should be updated.

Thanks

shumonsharif:
Hi @EmreKp One would only need to add the `cxf-rt-ws-security` dependency for CXF WS-Security support.  The `wss4j` dependencies should get pulled in automatically, given they are [compile time](https://mvnrepository.com/artifact/org.apache.cxf/cxf-rt-ws-security/3.5.0) dependencies.

There are examples of how to leverage the CXF WS-Security framework here:
https://github.com/apache/cxf/tree/3.5.x-fixes/distribution/src/main/release/samples/ws_security

Please feel free to submit a PR with the documentation improvement - spelling out the `cxf-rt-ws-security` dependency may possibly be helpful for others.

shumonsharif:
Something to note is that only the programmatic WSS4JInterceptors are currently supported.  Actions like Timestamp, UsernameToken, Signature, Encryption, etc., can be applied to the interceptors by passing the appropriate configuration properties.

EmreKp:
Thanks for reply, then I will submit PR with updated documentation.

PR #341, opened 2022-01-08
Add wss4j dependency and support info to related document (#340)

EmreKp:
Hello,

I updated documentation and submitted PR related with #340 . I also pasted the note for support of which interceptors, may be helpful for other developers.

Thanks..

### DIFF
--- a/docs/modules/ROOT/pages/ws-security.adoc
+++ b/docs/modules/ROOT/pages/ws-security.adoc
@@ -3,14 +3,13 @@
 
 The CXF framework's https://cxf.apache.org/docs/ws-security.html[WS-Security] implementation is based on https://ws.apache.org/wss4j/user_guide.html[WSS4J]. The example below shows how to integrate with WSS4J using interceptors.
 
-To use WSS4J interceptors in your code, firstly you should add CXF's WSS4J dependency (Replace `version` with CXF version used by Quarkus).
+To use WSS4J interceptors in your code, firstly you should add CXF's WS-Security dependency.
 
 [source,xml]
 ----
 <dependency>
       <groupId>org.apache.cxf</groupId>
       <artifactId>cxf-rt-ws-security</artifactId>
-      <version>{{cxf-version}}</version>
 </dependency>
 ----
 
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/ws-security.adoc
```diff
@@ -3,6 +3,19 @@
 
 The CXF framework's https://cxf.apache.org/docs/ws-security.html[WS-Security] implementation is based on https://ws.apache.org/wss4j/user_guide.html[WSS4J]. The example below shows how to integrate with WSS4J using interceptors.
 
+To use WSS4J interceptors in your code, firstly you should add CXF's WSS4J dependency (Replace `version` with CXF version used by Quarkus).
```
shumonsharif, line 6: - You are adding CXF's **WS-Security** dependency _(not WSS4J)_
- I don't believe we need to say _"(Replace `version` with CXF version used by Quarkus)"_ .... for example, if someone is using the CXF BOM, then they don't need a version specified here at all.  Would suggest we remove this statement, along with the version tag in the XML block.  A basic understanding of Maven can be expected of the extension users, so I think we can omit this level of detail.

shumonsharif (changes requested):
Thanks @EmreKp for this PR.  Would you kindly make the suggested updates?

docs/modules/ROOT/pages/ws-security.adoc
EmreKp: Thanks for suggestions, I changed version statement like "_you should add latest version_"  to satisfy any curiosities,  if it needs better statement, please let me know.

EmreKp:
Thanks for suggestions, I also changed version statement like "_you should add latest version_" to answer any questions, if it needs better statement, please let me know.

shumonsharif:
> Thanks for suggestions, I also changed version statement like "_you should add latest version_" to answer any questions, if it needs better statement, please let me know.

Thanks @EmreKp Yes, I would suggest we remove that statement all together; adding the latest version is not accurate either.

EmreKp:
> > Thanks for suggestions, I also changed version statement like "_you should add latest version_" to answer any questions, if it needs better statement, please let me know.
> 
> Thanks @EmreKp Yes, I would suggest we remove that statement all together; adding the latest version is not accurate either.

I removed it for now, maybe it can be added later if some users would have problem concerned with that.

Issue #346, opened 2022-01-24
Native build using @RolesAllowed, quarkus-security and quarkus-cxf


odellamaggiore:
Given a simple app (built from quarkus-getting started) using : 
- quarkus-cxf
- quarkus-security
- @RolesAllowed annotation on a soap endpoint

Soap Endpoint: 
```
@WebService(endpointInterface = "org.acme.getting.started.GreetingSoap")
@RolesAllowed("SpecialRole")
public class GreetingSoapResource implements GreetingSoap {

    @Override
    public String sayHello(String name) {
        return "Hello " + name;
    }
}
```

Dependencies:
```
<dependency>
  <groupId>io.quarkus</groupId>
  <artifactId>quarkus-resteasy</artifactId>
</dependency>
<dependency>
  <groupId>io.quarkiverse.cxf</groupId>
  <artifactId>quarkus-cxf</artifactId>
  <version>0.12.0</version>
</dependency>
<dependency>
  <groupId>io.quarkus</groupId>
  <artifactId>quarkus-security</artifactId>
</dependency>
```

In native mode, the application fails to start : 

```
2022-01-24 10:51:59,123 INFO  [org.apa.cxf.wsd.ser.fac.ReflectionServiceFactoryBean] (main) Creating Service {http://started.getting.acme.org/}GreetingSoapResource_SubclassService from class org.acme.getting.started.GreetingSoap
2022-01-24 10:51:59,129 ERROR [io.qua.run.Application] (main) Failed to start application (with profile prod): java.lang.NoSuchMethodException: org.acme.getting.started.GreetingSoapResource_Subclass.sayHello(java.lang.String)
        at java.lang.Class.getMethod(DynamicHub.java:1150)
        at org.apache.cxf.jaxws.JAXWSMethodDispatcher.getImplementationMethod(JAXWSMethodDispatcher.java:89)
```
If we remove the @RolesAllowed annotation, or the quarkus-security import, it works fine.
This issue appends only in native mode.

Example project to reproduce : 
[cxf-rolesallowed-issue.zip](https://github.com/quarkiverse/quarkus-cxf/files/7925067/cxf-rolesallowed-issue.zip)

shumonsharif:
Hi @odellamaggiore A couple questions for you.

[...]

PR #347, opened 2022-01-26
Address native build issue for security annotations

shumonsharif:
This addresses #346 ... looking at this very much as a temporary workaround until https://github.com/quarkusio/quarkus/issues/7507 is fixed.

@famod Would you please review?  I'm still undecided if this PR worth the trouble, or whether we should just hold off for the Quarkus ArC fix - though unsure if / when that may be.  Your thoughts would be much appreciated.

### DIFF
--- a/deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
+++ b/deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
@@ -283,7 +283,7 @@ class QuarkusCxfProcessor {
                      * https://github.com/quarkiverse/quarkus-cxf/issues/346
                      * https://github.com/quarkusio/quarkus/issues/7507
                      */
-                    reflectiveClass.produce(new ReflectiveClassBuildItem(true, true, impl + "_Subclass"));
+                    reflectiveClass.produce(new ReflectiveClassBuildItem(true, false, impl + "_Subclass"));
 
                     String soapBinding = Optional.ofNullable(wsClass.classAnnotation(BINDING_TYPE_ANNOTATION))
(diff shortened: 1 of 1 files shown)


deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
```diff
@@ -277,6 +277,14 @@ public void build(
                     String impl = wsClass.name().toString();
                     String wsName = impl.contains(".") ? impl.substring(impl.lastIndexOf('.') + 1) : impl;
                     additionalBeans.produce(AdditionalBeanBuildItem.unremovableOf(impl));
+
+                    /*
+                     * Address issue #346; though this really should be handled in ArC
+                     * https://github.com/quarkiverse/quarkus-cxf/issues/346
+                     * https://github.com/quarkusio/quarkus/issues/7507
+                     */
+                    reflectiveClass.produce(new ReflectiveClassBuildItem(true, true, impl + "_Subclass"));
```
famod, line 286: Do we really need fields or are methods enough?

famod (approved):
If it fixes an issue for a user without any known regressions I'm all for it.
I have only one small remark.

PS: Funny, this revealed a (not really related) issue in Quarkus: #23229

deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
shumonsharif: I thought we may need both from looking at `SubclassGenerator` - but now that you asked, I tested it and actually seems we just need methods.  Fixed in latest commit.

Issue #345, opened 2022-01-20
Invalid URL generated in WSDL for included XSD files


seckinsq:
We are having a similar issue as #323 but we are using our own WSDL file and we use some include and imports in WSDL or XSD files. When we want to retrieve wsdl from service (?wsdl), the URL used for _schemaLocation_ on _xs:include_ is not valid.

`<xs:include schemaLocation="http://localhost:8080/soap/CxftestSoapService?wsdl?xsd=cxftest.xsd"/>`

It seems like application is adding path query for XSD to WSDL url instead of service URL. 

http://localhost:8080/soap/CxftestSoapService?wsdl?xsd=cxftest.xsd 
(Gives soap:Server No binding operation info while invoking unknown method with params unknown.)

vs 

http://localhost:8080/soap/CxftestSoapService?xsd=cxftest.xsd 
(This works but if it imports another XSD, its query path is added to this XSD's url  too : CxftestSoapService?xsd=cxftest.xsd?xsd=common.xsd)

Example mini-project to reproduce:
[code-with-quarkus.zip](https://github.com/quarkiverse/quarkus-cxf/files/7906988/code-with-quarkus.zip)

shumonsharif:
Hi @seckinsq Thanks for the issue.  Did you by any chance try exploring the `published-endpoint-url` property as a workaround?  I believe you should be able to set that value to either an _absolute_ or _relative_ URL ... can you please try and let us know your findings?

seckinsq:
Hi, @shumonsharif thank you for fast response. 

We tried absolute URL and it was OK but we cannot use it as there are many environments where this application will be running. All will have different host (and there may be different versions)

I have tried your suggestion with relative URL : 
`quarkus.cxf.endpoint."CxftestSoapService".published-endpoint-url=/soap/CxftestSoapService` 

For http://localhost:8080/soap/CxftestSoapService?wsdl
`<xs:include schemaLocation="/soap/CxftestSoapService?xsd=cxftest.xsd"/>`

For http://localhost:8080/soap/CxftestSoapService?xsd=cxftest.xsd
`<xs:import schemaLocation="/soap/CxftestSoapService?xsd=common.xsd" namespace="http://cxftest.com/common/schema"/>`

So they are looking good with relative URL, I will check with client applications if this can be acceptable.
Thanks.

PR #349, opened 2022-01-28
Fix invalid import URLs in WSDLs

### LINKED ISSUE #323: Invalid URL generated in XML "import"
Starting from 0.10.0, when publishing a webservice with the _targetNamespace_ attribute of the `@WebService` annotation, XML schema is not embedded within the WSDL but linked via _import_; the problem is that the _import_ statement contains an invalid URL, such as this one:
`<wsdl:import location="http://localhost:8080/soap/test?wsdl?wsdl=TestService.wsdl" namespace="mynamespace"> </wsdl:import>`

Workarounds:
- Don't specify a targetNamespace (but some services may need it);
- Include a pre-generated WSDL (but the WSDL file would need to be manually edited whenever a field is added to the model, for example);
- Specify a _quarkus.cxf.endpoint.key.published-endpoint-url_ property (not optimal because different environments would need different configurations; in most use cases auto-detection of the published URL is desirable).

Example mini-project to reproduce: [cxftest.zip](https://github.com/quarkiverse/quarkus-cxf/files/7642790/cxftest.zip)

shumonsharif:
Fixes #323 and #345.

@famod Please have a quick review.  The core issue was that `getRequestURL()` was returning the URL with query parameters in it; it should have been without.  This was a tricky one to figure out.

This not only fixes the issues with the XSD imports, but also the service address location should appear correctly in the WSDLs now, as shown below, without any query parameters (eg. ?wsdl) in it.

```xml
<wsdl:service name="CxftestSoapService">
    <wsdl:port binding="tns:CxftestSoapBinding" name="CxftestServicePort">
        <soap:address location="http://localhost:8080/soap/CxftestSoapService"/>
    </wsdl:port>
</wsdl:service>
```

### DIFF
--- a/runtime/src/main/java/io/quarkiverse/cxf/transport/VertxHttpServletRequest.java
+++ b/runtime/src/main/java/io/quarkiverse/cxf/transport/VertxHttpServletRequest.java
@@ -449,7 +449,9 @@ public class VertxHttpServletRequest implements HttpServletRequest {
 
     @Override
     public StringBuffer getRequestURL() {
-        return new StringBuffer(request.absoluteURI());
+        String absoluteUri = request.absoluteURI();
+        String urlWithoutParams = absoluteUri.replaceFirst("\\?.*$", "");
+        return new StringBuffer(urlWithoutParams);
     }
 
     @Override
--- a/integration-tests/src/test/java/io/quarkiverse/it/cxf/XForwardedHeadersTest.java
+++ b/integration-tests/src/test/java/io/quarkiverse/it/cxf/XForwardedHeadersTest.java
@@ -40,7 +40,7 @@ public class XForwardedHeadersTest {
                 .get("/soap/greeting?wsdl")
                 .then()
                 .statusCode(200)
-                .body(containsString("http://localhost:8081/test/soap/greeting?wsdl"));
+                .body(containsString("http://localhost:8081/test/soap/greeting"));
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


shumonsharif:
@famod I am pushing this one through for now, so I can get a new release out.  But please feel to review the updates at your convenience and let me know if anything catches your eye, we can always revisit the code as needed!

famod:
I'm having a look later or tomorrow.
Btw, I want to silence some INFO logs but I haven't found the time to do so yet. But I guess that'll have to wait for the next release then.

runtime/src/main/java/io/quarkiverse/cxf/transport/VertxHttpServletRequest.java
```diff
@@ -449,7 +449,9 @@ public String getRequestURI() {
 
     @Override
     public StringBuffer getRequestURL() {
-        return new StringBuffer(request.absoluteURI());
+        String absoluteUri = request.absoluteURI();
+        String urlWithoutParams = absoluteUri.replaceFirst("\\?.*$", "");
```
famod, line 453: Do we know how often this method is called per "typical" request?
I'm asking because here a thrown-away `Pattern` is created under the hood, so that each invocation will go through a re-compilation the `Pattern`.
I think the `Pattern` should be pre-compiled once (as a constant).

shumonsharif: Nice catch Falko!  Each request invocation will have a call to this method, so you're quite right, there will be a performance hit here.  I think we can do a "0.13.1" release with your logging updates and this change?

famod: Sounds good! I'll have a look at the logging stuff tomorrow.

shumonsharif: Sounds good, thanks.  New PR submitted:  https://github.com/quarkiverse/quarkus-cxf/pull/351

PR #351, opened 2022-01-29
Performance enhancement for PR #349

shumonsharif:
@famod This should address the performance impacts of using the previous regex based approach.  I got rid of the `Pattern` dependency all together, this may be a slightly more performant approach.


famod (approved):
👍

PR #363, opened 2022-02-19
Bug fix and log cleanup

shumonsharif:


shumonsharif:
Woops just realized I forgot to request a review here too ... @famod

runtime/src/main/java/io/quarkiverse/cxf/graal/CXFSubstitutions.java
```diff
@@ -76,7 +76,7 @@ public WrapperHelper compile(Class<?> wrapperType, Method[] setMethods,
             Method[] getMethods, Method[] jaxbMethods,
             Field[] fields, Object objectFactory) {
         Logger LOG = LogUtils.getL7dLogger(Target_org_apache_cxf_jaxb_WrapperHelperClassGenerator.class);
-        LOG.info("compileWrapperHelper substitution");
+        LOG.fine("compileWrapperHelper substitution");
```
famod, line 79: I think `debug` is more common. But it's a matter of taste...

runtime/src/main/java/io/quarkiverse/cxf/graal/CXFSubstitutions.java
```diff
@@ -100,7 +100,7 @@ public WrapperHelper compile(Class<?> wrapperType, Method[] setMethods,
                     try {
                         cls = Thread.currentThread().getContextClassLoader().loadClass(newClassName);
                     } catch (ClassNotFoundException e) {
-                        LOG.warning("Wrapper helper class not found : " + e.toString());
+                        LOG.fine("Wrapper helper class not found : " + e.toString());
```
famod, line 103: I missed that myself elsewhere (need to send a PR for that some day), but we should use the "formatting" variant wherever possible, meaning `.debugf("Wrapper helper class not found : %s", e)`.

runtime/src/main/java/io/quarkiverse/cxf/graal/CXFSubstitutions.java
shumonsharif: This one is actually not a `org.jboss.logging.Logger`, instead it is [`java.util.logging.Logger`](https://docs.oracle.com/en/java/javase/11/docs/api/java.logging/java/util/logging/Logger.html) ... there is no debug or debugf method.

runtime/src/main/java/io/quarkiverse/cxf/graal/CXFSubstitutions.java
shumonsharif: Agreed, I think we need a general PR to address much of the logging throughout the code; the intent with this PR was just the minor log level change.

runtime/src/main/java/io/quarkiverse/cxf/graal/CXFSubstitutions.java
famod: Ah, good point.

Issue #354, opened 2022-01-31
CXF WS-Security native mode support


wanderingfella:
Hi,

I am having an issue using ws security in native mode.
I am using 
`'io.quarkiverse.cxf:quarkus-cxf:0.13.0'`
and also added
`'org.apache.cxf:cxf-rt-ws-security:3.4.4'`
I also had to add 
`com.sun.xml.messaging.saaj:saaj-impl:1.5.3`

I think to make it work in JVM mode.

It works fine in JVM mode, but having an issue when running it in native mode.

The top of the stack trace is below

`Caused by: org.apache.cxf.binding.soap.SoapFault: Problem writing SAAJ model to stream: Unable to create message factory for SOAP: Unable to create SAAJ meta-factory: Provider com.sun.xml.internal.messaging.saaj.soap.SAAJMetaFactoryImpl not found\n\tat org.apache.cxf.binding.soap.saaj.SAAJOutInterceptor.handleMessage(SAAJOutInterceptor.java:140)\n\tat org.apache.cxf.ws.security.wss4j.WSS4JOutInterceptor.handleMessage(WSS4JOutInterceptor.java:102)\n\tat org.apache.cxf.ws.security.wss4j.WSS4JOutInterceptor.handleMessage(WSS4JOutInterceptor.java:58`

I think as it's not picking up META-INF/services inside the saaj impl jar.

Which seems to define com.sun.xml.messaging.saaj.soap.SAAJMetaFactoryImpl as being the factory impl.

In native mode it seems to be looking for com.sun.xml.**internal.**messaging.saaj.soap.SAAJMetaFactoryImpl.

Not really sure how to work around this, it seems like I need to register the saaj impl jar for reflection and the META-INF/service.
I hope that makes sense, any suggestions appreciated.

shumonsharif:
> I am having an issue using ws security in native mode. I am using `'io.quarkiverse.cxf:quarkus-cxf:0.13.0'` and also added `'org.apache.cxf:cxf-rt-ws-security:3.4.4'` I also had to add `com.sun.xml.messaging.saaj:saaj-impl:1.5.3`
> 
> I think to make it work in JVM mode.

Hi @wanderingfella The SAAJ (SOAP with Attachments API for Java) - as the name indicates - is used for attachments and may possibly not be required for WS-Security.  If it is indeed required, then the `cxf-rt-ws-security` dependency I think should pull it in.

[...]

PR #364, opened 2022-02-19
WS-Security native mode support

shumonsharif:
This should address #354 ; docs have been updated to provide additional details regarding WS-Security native mode support.

@famod Would you kindly review when you have some time?  Thanks in advance!

### DIFF
--- a/deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfWsSecurityProcessor.java
+++ b/deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfWsSecurityProcessor.java
@@ -1,0 +1,332 @@
+package io.quarkiverse.cxf.deployment;
+
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
+import java.net.URL;
+import java.util.Arrays;
+import java.util.List;
+
+import io.quarkiverse.cxf.graal.IsWsSecurityPresent;
+import org.jboss.logging.Logger;
+
+import io.quarkus.deployment.annotations.BuildProducer;
+import io.quarkus.deployment.annotations.BuildStep;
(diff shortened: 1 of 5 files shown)


shumonsharif:
BTW @famod I couldn't find a good way to deal with the `--allow-incomplete-classpath` flag for two optional transitive dependencies that were missing from the classpath during the native build:

- `org.jvnet.mimepull:mimepull`
- `org.apache.cxf:cxf-rt-ws-mex`

I tried the [NativeImageAllowIncompleteClasspathBuildItem](https://github.com/quarkusio/quarkus/blob/main/core/deployment/src/main/java/io/quarkus/deployment/builditem/nativeimage/NativeImageAllowIncompleteClasspathBuildItem.java), but it not only printed a stark warning, it also did not seem like a good option anyway.

I also couldn't find any properties (Maven or Quarkus) that would force the inclusion of optional dependencies - in the end, I just updated the docs to say that one would need to include the two optional dependencies in their WS-Security project for the native build to work.

famod:
Just saw this. I'll try to have a look later this week.

shumonsharif:
A few notes:

- I was trying to avoid introducing a dependency on `cxf-rt-ws-security` for the extension.
- I had strongly considered splitting up the codebase to introduce sub-modules for [each of the `cxf-rt-ws-*`](https://github.com/apache/cxf/tree/master/rt/ws) dependencies.  While it would simplify and result in significantly cleaner code, it'll be a major refactoring and involved effort.  I feel it is the better approach, and will keep looking into it, though my bandwidth is limited.
- There is some odd behavior with the GraalvM native image and locales that I could not quite get to the bottom of (I'd need to debug the native image), which resulted in the introduction of this [somewhat odd method](https://github.com/quarkiverse/quarkus-cxf/pull/364/files#diff-6d93c9f954f11ebbae9fbae363f3d64e4862b54dc79f44df3ffd257574e7b3a5R276).  Basically, the properties defined in wss4j_errors.properties were not available as a resource bundle (tried numerous approaches to address the issue); in the end I merged the properties into a bundle that is.

shumonsharif:
@famod I'm thinking of merging this in for now; let me know in case in case I should hold off, or you can always do a retroactive review once you have some time - and will make any corrections at that point in time.

PR #366, opened 2022-02-22
Build in other OSes too

gastaldi:
This PR changes the build script to build in Windows too

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Extract GraalVM version
         id: graalvm
         run: |
-          echo "::set-output name=version::$(mvn -ntp dependency:list -DincludeArtifactIds=graal-sdk | grep graal-sdk | cut -d: -f4)"
+          echo "::set-output name=version::$(mvn -pl runtime -ntp dependency:list -DincludeArtifactIds=graal-sdk | grep graal-sdk | cut -d: -f4)"
           
       - name: Setup GraalVM ${{ steps.graalvm.outputs.version }}
         uses: graalvm/setup-graalvm@v1
@@ -59,7 +59,7 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Build - JVM
-        run: mvn -B formatter:validate install --file pom.xml
+        run: mvn -B formatter:validate verify --file pom.xml
         
       - name: Build - Native
         run: mvn -B verify -Pnative --file integration-tests/pom.xml
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,13 +37,18 @@ jobs:
 
       - uses: actions/checkout@v2
 
-      - name: Cache local Maven repository
+      - name: Get Date
+        id: get-date
+        run: |
+          echo "::set-output name=date::$(/bin/date -u "+%Y-%m")"
+        shell: bash
+      - name: Cache Maven Repository
+        id: cache-maven
         uses: actions/cache@v2
         with:
           path: ~/.m2/repository
-          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
-          restore-keys: |
-            ${{ runner.os }}-maven-
+          # refresh cache every month to avoid unlimited growth
+          key: maven-repo-${{ runner.os }}-${{ steps.get-date.outputs.date }}
 
       - name: Extract GraalVM version
         id: graalvm
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,8 +58,5 @@ jobs:
           components: 'native-image'
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Build - JVM
-        run: mvn -B formatter:validate verify --file pom.xml
-        
-      - name: Build - Native
-        run: mvn -B verify -Pnative --file integration-tests/pom.xml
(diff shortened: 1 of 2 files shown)


gastaldi:
Ah darn, this probably won't work because Quarkus does not build in Docker by default in Windows.

shumonsharif:
@gastaldi This would be a good PR if it works!  Is there no way to run the `gu install native-image` command in the VM?

gastaldi:
Maybe, or perhaps https://github.com/marketplace/actions/graalvm-native-image-toolchain

famod:
The latter is archived.

Isn't that the "official" one? https://github.com/marketplace/actions/github-action-for-graalvm

gastaldi:
@famod ah that looks much better, let's see how that behaves

famod:
So with this change also the JVM mode is running on GraalVM, right? Not the most commons setup, or is it?
I do understand though that it's easier this way...

shumonsharif:
So the Windows native build is actually NOT passing, even though it may appear that way ...

Several warnings in the logs (that are not there in the linux build):
https://github.com/quarkiverse/quarkus-cxf/runs/5295906918?check_suite_focus=true#step:7:241

Followed by what looks like a port connection getting terminated:
https://github.com/quarkiverse/quarkus-cxf/runs/5295906918?check_suite_focus=true#step:7:337

famod:
> Several warnings in the logs (that are not there in the linux build): https://github.com/quarkiverse/quarkus-cxf/runs/5295906918?check_suite_focus=true#step:7:241

You can see the same warning in the Quarkus CI: https://github.com/quarkusio/quarkus/runs/5285752724?check_suite_focus=true#step:12:122
The subsequent warnings are a consequence of the uber jar mode.
AFAICS, this is all "expected", as unpretty as it looks...

> Followed by what looks like a port connection getting terminated: https://github.com/quarkiverse/quarkus-cxf/runs/5295906918?check_suite_focus=true#step:7:337

Doesn't seem to fail the test though. I'm not saying it isn't a problem, but it seems to happen at the very end when things a torn down.

shumonsharif:
Looking deeper, the warnings appear to all be uberjar related, and can possibly be ignored.  Also, the connection terminated looks more like an unclean shutdown of the native process, and all the unit tests actually passed.  So I think we may be OK.

gastaldi:
> So with this change also the JVM mode is running on GraalVM, right? Not the most commons setup, or is it? I do understand though that it's easier this way...

Yeah, we could consider running on Mandrel too, might need to experiment a little more.

shumonsharif:
> > So with this change also the JVM mode is running on GraalVM, right? Not the most commons setup, or is it? I do understand though that it's easier this way...
> 
> Yeah, we could consider running on Mandrel too, might need to experiment a little more.

Isn't Mandrel a GraalVM based distribution?  Is that what was meant, or was it more just using OpenJDK for JVM mode?

shumonsharif (approved):

famod:
> or was it more just using OpenJDK for JVM mode?

This; Temurin, Zulu or Liberica. Doesn't matter much in the end, I suppose.

shumonsharif:
@gastaldi Are we good to merge this?

gastaldi:
My only concern is that the GraalVM used to build the native image might not be the same used in Quarkus, maybe that should be a fixed value and need to be updated manually when a new Quarkus version comes up

gastaldi:
Yeah, I think that could probably work

gastaldi:
Almost worked :)

shumonsharif:
Not sure how it's ending up with the repeating version number here:  
https://github.com/quarkiverse/quarkus-cxf/runs/5328047364?check_suite_focus=true#step:6:3 

Is it somehow reusing the the value from the linux flow, which also seems to repeat?
https://github.com/quarkiverse/quarkus-cxf/runs/5328047342?check_suite_focus=true#step:5:4

gastaldi:
@shumonsharif That's because the dependency:list is run in the multimodule project. Let me see if I can bypass that

.github/workflows/build.yml
```diff
@@ -59,7 +59,7 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Build - JVM
-        run: mvn -B formatter:validate install --file pom.xml
+        run: mvn -B formatter:validate verify --file pom.xml
```
shumonsharif, line 62: @gastaldi This will likely fail in future builds; `install` puts the quarkus-cxf artifacts into the local maven repo, so that they are available for the next step (the native build), and `verify` I believe doesn't do that.  I'm guessing the build passed for now because you set up a cache which coincidentally already had the necessary artifacts.

.github/workflows/build.yml
```diff
@@ -59,7 +59,7 @@ jobs:
...
         
       - name: Build - Native
         run: mvn -B verify -Pnative --file integration-tests/pom.xml
```
shumonsharif, line 65: This one can remain as `verify`, but it depends on artifacts as mentioned from the previous step.

shumonsharif (changes requested):

.github/workflows/build.yml
gastaldi: Right, because JVM tests were executed in the second step anyway I merged them in a single verify call

shumonsharif:
@gastaldi is it ok with you if I take a look at this and commit to your branch?

gastaldi:
Sure, I'll test separately and provide a new PR once I get it working

PR #379, opened 2022-03-16
Modularization, JAX-WS Handlers integration and Documentation updates

shumonsharif:
This will address #375 

@dufoli @famod Not sure whether you both are tracking these notifications, but I plan on merging this within the next couple days if there are no objections.


famod:
Sorry, late to the party. I think this makes a lot of sense, given that cxf itself has many modules. 👍

shumonsharif:
Good to hear from you Falko, and better late than never :)

If you have the bandwidth, please feel free to do a retroactive review of the PR, and I'll be happy to address any issues!

PR #466, opened 2022-07-14
Introduce quarkus-cxf-bom

### LINKED ISSUE #460: Provide a BOM?
First off, I haven't checked with others, especially @aloubyansky or @gastaldi, whether this is already solved in some general way for quarkiverse extensions or is being worked at.

In my current project we are using `org.apache.cxf:cxf-rt-transports-http-netty-server` to fire up a CXF test server.
Since `quarkus-cxf` is not depending on that module, we have to define the version for that module on our own **and we need to take care to keep it in sync with the version of cxf that `quarkus-cxf` brings in.**

To avoid that, what about providing a `quarkus-cxf-bom` that imports [`cxf-bom`](https://mvnrepository.com/artifact/org.apache.cxf/cxf-bom/3.5.2)?

famod:
Resolves #460

A few remarks:

It's a shame Maven doesn't support global exclusions or at least BOM exclusions ([MNG-5600](https://issues.apache.org/jira/browse/MNG-5600)), so we have to re-declare certain dependencies from `cxf-bom` to define the required exclusions.
The potential issue with that is that if someone uses this new bom and picks a module that we don't cover in our BOM in terms of exclusions, than something like `geronimo-jta_1.1_spec` can still end up on the classpath.
I don't see how to get around this, _except_ listing all cxf deps, which defeats the idea of importing `cxf-bom` in the first place.
Maybe we should add something to the docs about this?

I was able to remove the explicit dependency to `jakarta.xml.ws-api` without breaking anything (AFAICS). This now comes in transitively via e.g. `cxf-rt-frontend-jaxws`. I might be missing something here, though.
I compared dependency trees before and after this change in core and security modules and the only change was in `core/deployment`:
Before:
```
 +- jakarta.xml.ws:jakarta.xml.ws-api:jar:2.3.3:compile
 |  +- jakarta.xml.bind:jakarta.xml.bind-api:jar:2.3.3:compile
 |  |  \- jakarta.activation:jakarta.activation-api:jar:1.2.1:compile
 |  +- jakarta.xml.soap:jakarta.xml.soap-api:jar:1.4.2:compile
 |  \- jakarta.jws:jakarta.jws-api:jar:2.1.0:compile
```
After:
```
 +- io.quarkiverse.cxf:quarkus-cxf:jar:1.3.0-SNAPSHOT:compile
 |  +- ...
 |  +- org.apache.cxf:cxf-rt-frontend-jaxws:jar:3.5.3:compile
 |  |  +- ...
 |  |  +- jakarta.xml.ws:jakarta.xml.ws-api:jar:2.3.3:compile
 |  |  +- jakarta.jws:jakarta.jws-api:jar:2.1.0:compile
 |  |  +- jakarta.xml.soap:jakarta.xml.soap-api:jar:1.4.2:compile
 |  |  |  \- jakarta.activation:jakarta.activation-api:jar:1.2.1:compile
```


famod:
@aloubyansky @gastaldi you might be interested in this as well. It's a "simpler" setup than in the Quarkus main repo, given that we don't have a `build-parent` here. So we need to be careful with adding any `dependencyManagement` to the root `quarkus-cxf-parent` because that would be inherited and propagated by the new BOM (fell into that trap just recently in another project).

bom/pom.xml
```diff
@@ -0,0 +1,158 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkiverse.cxf</groupId>
+        <artifactId>quarkus-cxf-parent</artifactId>
+        <version>1.3.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
```
gastaldi, line 9: BOMs usually don't declare a parent, but I can see why this is needed

famod: Yeah, this is related to that inheritance pitfall I mentioned further up.

Without it distributionManagement etc. is missing.

gastaldi: Maybe it would be interesting to consider using https://www.mojohaus.org/flatten-maven-plugin/ when releasing

bom/pom.xml
```diff
@@ -0,0 +1,158 @@
...
+            </dependency>
+
+            <dependency>
+                <groupId>org.apache.cxf</groupId>
+                <artifactId>cxf-rt-ws-mex</artifactId>
+                <version>${cxf.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>org.apache.cxf</groupId>
+                        <artifactId>cxf-core</artifactId>
```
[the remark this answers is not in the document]
shumonsharif: @famod Please question everything.  :)  This was probably a remnant of trying to get WS-Security native mode in a working state.  This needs to be cleaned up a little, so would suggest we remove core and the other two deps from exclusion.

shumonsharif (approved):
@famod Looks great, just had the one minor comment (unrelated to your changes) about cleaning up some of the unnecessary exclusions such as core.  If you like, I can take that up separately, just let me know.

shumonsharif:
@famod I'm merging your PR for now; I'll take care of the core and other dependency exclusions cleanup separately.

> It's a shame Maven doesn't support global exclusions or at least BOM exclusions ([MNG-5600](https://issues.apache.org/jira/browse/MNG-5600)), so we have to re-declare certain dependencies from `cxf-bom` to define the required exclusions. The potential issue with that is that if someone uses this new bom and picks a module that we don't cover in our BOM in terms of exclusions, than something like `geronimo-jta_1.1_spec` can still end up on the classpath. I don't see how to get around this, _except_ listing all cxf deps, which defeats the idea of importing `cxf-bom` in the first place. Maybe we should add something to the docs about this?

Regarding the above comment, updating the docs will be good; I'll take a stab at that as well and keep you posted.

famod:
@shumonsharif sorry for the lack if feedback! I'm fine with all you've said. 😄

PR #467, opened 2022-07-14
Set Quarkus 2.10.2.Final in docs

famod:
I can't say I'm familiar with how exactly docs are generated/built, but I kept getting this change locally, so I suppose this was forgotten somehow.

### DIFF
--- a/docs/modules/ROOT/pages/attributes.adoc
+++ b/docs/modules/ROOT/pages/attributes.adoc
@@ -1,2 +1,2 @@
-:quarkus-version: 2.10.0.Final
+:quarkus-version: 2.10.2.Final
 :quarkus-cxf-version: 1.2.0


shumonsharif:
Hey @famod That file is used for the documentation; I was trying to auto generate and auto update the version numbers during the build and release process, as shown here:
https://github.com/quarkiverse/quarkus-cxf/blob/master/.github/workflows/release.yml#L54-L57

I believe adding the `docs/modules/ROOT/pages/attributes.adoc` file to `.gitignore` and forcing the `git add` with `-f` in the snippet above should do the trick, though I didn't get around to testing it.  Please feel free to make these minor updates, or let me know if you prefer I take it up!

shumonsharif:
@famod I just created and merged https://github.com/quarkiverse/quarkus-cxf/pull/468 which I _**thought**_ may address this issue.  Upon further testing, and trying a whole bunch of different approaches, I finally came to [this conclusion](https://git-scm.com/docs/gitfaq#ignore-tracked-files), the hard way :anguished: - it's not possible to ignore changes to a tracked file, at least in a generic and seamless way where a fresh clone would be set up properly.

Closest thing I could find was running the below locally to ignore the local changes, but this also appears to have some unwanted side effects:
`git update-index --skip-worktree docs/modules/ROOT/pages/attributes.adoc`

Not sure how to handle this cleanly just yet ... for now, I'm merging your PR in, but keep in mind the `docs/modules/ROOT/pages/attributes.adoc` file will get auto generated and committed during each release with whatever happens to be in the version numbers for that release.

famod:
@shumonsharif 
I might be missing something, but couldn't the docs module filter files to `target`? I think that would be more like the "standard maven way".
Alternatively, `asciidoctor-maven-plugin` has built-in `attributes` support: https://docs.asciidoctor.org/maven-tools/latest/plugin/goals/process-asciidoc/#configuration

shumonsharif:
@famod The issue is not quite related to Maven.

The way it works is that the [quarkiverse-docs](https://github.com/quarkiverse/quarkiverse-docs) project uses [Antora](https://antora.org/) to generate the docs - Maven is not in play at all during the website generation process.  

It pulls down the code from each Quarkiverse repo (`target` or `asciidoctor-maven-plugin` is not in the picture at all), and runs an Antora [playbook](https://github.com/quarkiverse/quarkiverse-docs/blob/main/antora-playbook.yml) to build the docs [website](https://quarkiverse.github.io/quarkiverse-docs/index/index/index.html).

famod:
I see, so it needs "persisted" attributes.

Other ideas:
- also push to master
- run respective plugin only in release profile

shumonsharif:
Thanks @famod - can't believe I didn't think of this!

> * run respective plugin only in release profile

I'm not entirely sure how maven profile inheritance and interpolation works, given the release profile is already defined in `quarkiverse-parent` and now will also be in the `docs` submodule.  If the interpolation is an issue, we can always create a unique profile just for the plugin and update the release pipeline appropriately ... in fact, I think the latter may be cleaner as it's all isolated in the extension.  Will give it a shot and keep you posted, and thanks again for the suggestion!

famod:
@shumonsharif any updates? Would be nice to have a new release of quarkus-cxf with the new BOM.
IMO that profile could be added later.

shumonsharif:
Hey @famod Sorry I've been MIA, been tied up with a few things!  I've been meaning to get to everything this weekend, will that work for you?  Let me know if not, I'll get wrapped up this evening if that helps.

famod:
Hi there! No stress, next week is fine. I'm on PTO for a few weeks beginning tomorrow anyway, so it's up to my colleagues to reap what I've sown. 😁

Issue #485, opened 2022-08-09
Issue with SecureRandom in org.apache.xml.security.stax.ext.XMLSecurityConstants


javaduke:
We are facing an issue with the WS-Security, the `nonce` is not added to the request even though we set `addUsernameTokenNonce` to true. Apparently the reason is that there's an exception due to the non-initialized secure random in `org.apache.xml.security.stax.ext.XMLSecurityConstants`. I saw that ws-security module substitutes the XMLSecurityConstants class, but the SecureRandom is still initialized at the build time, which causes the problem. Not sure how to address this.
See the https://github.com/apache/camel-quarkus/issues/3966 , in particular, Peters comment here: https://github.com/apache/camel-quarkus/issues/3966#issuecomment-1209360984
Wondering if anyone can give us a hand with this.

Many thanks!

ppalaga:
This should not be an issue anymore as of Quarkiverse CXF 1.5.0

PR #486, opened 2022-08-09
Initializing SECURE_RANDOM at the runtime

javaduke:
I don't really know if this is a right thing to do but it resolves the problem described here https://github.com/quarkiverse/quarkus-cxf/issues/485 and https://github.com/apache/camel-quarkus/issues/3966

### DIFF
--- a/ws-security/runtime/src/main/java/io/quarkiverse/cxf/graal/CxfWsSecuritySubstitutions.java
+++ b/ws-security/runtime/src/main/java/io/quarkiverse/cxf/graal/CxfWsSecuritySubstitutions.java
@@ -12,6 +12,7 @@ import javax.xml.datatype.DatatypeFactory;
 import javax.xml.stream.XMLOutputFactory;
 
 import org.apache.wss4j.stax.setup.WSSec;
+import org.apache.xml.security.exceptions.XMLSecurityException;
 import org.apache.xml.security.utils.Constants;
 
 import com.oracle.svm.core.annotate.Alias;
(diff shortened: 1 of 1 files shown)


shumonsharif:
Hi @javaduke Thanks for the PR!  This looks perfectly acceptable to me.  The intent with the original code was to switch the SECURE_RANDOM initialization by moving it from the static initializer block into the `XMLSecurityConstants` constructor.  Your PR moves it down to the `generateBytes` method, which should also be fine.

When do you need this to be released?

javaduke:
Awesome, thank you very much!!! It would be nice to release it soon, but I guess we can wait a few days, let me ask when is the next Camel Quarkus release is planned and I'll let you know.

ppalaga:
Thanks @javaduke! It's 7 days to 2.12.0.CR1 https://github.com/quarkusio/quarkus/wiki/Release-Planning#212-proposal

shumonsharif:
> I'd like to check whether we could avoid the substitutions in quarkus-cxf somehow. Substitutions tend to be fragile and hard to maintain in the long term.

I completely agree that we should avoid substitutions if at all possible, but this was the only solution I could come up with, given XMLSecurityConstants initializes SECURE_RANDOM in it's static initializer block.
https://github.com/apache/santuario-xml-security-java/blob/2.3.x-fixes/src/main/java/org/apache/xml/security/stax/ext/XMLSecurityConstants.java#L51

Open to suggestions though, in case there may be a better way to handle this.  I will drop a release tomorrow, but let me know in case I should hold off.

ppalaga:
Hold off please. I am working on some changes.

PR #488, opened 2022-08-13
Multiple test modules

ppalaga:
The ultimate goal is to have multiple test modules which would separately cover various aspects of Quarkus CXF:

* WS Server (done, but the module is still called `basic`)
* WS Client (done)
* WS Security (TODO: https://github.com/quarkiverse/quarkus-cxf/issues/498 )

and possibly more. Having those in separate Maven modules is important because each use case may require different set of dependencies.


ppalaga:
If somebody could please pre-approve this so that I can see that my changes work on the CI.

shumonsharif:
Hey @ppalaga So I attempted to get you added to the project on #489 - but it didn't work.  I believe we'll need at least one successfully merged PR from you, before that'll kick in, or maybe it's a permission issue.

@gastaldi Any thoughts on this?

gastaldi:
@shumonsharif I noticed that it doesn't work if the comment is the first in the issue. You can add a new comment there with the same text and the all contributors bot should be triggered

gastaldi:
BTW you can use that comment anywhere (even here), no need to create a specific issue

gastaldi:
Also adding as a contributor with @all-contributors doesn't warrant CI runs, you're right that a PR needs to be merged  to bypass that

shumonsharif:
Thanks @gastaldi for your help!  Just got home from a long walk, here goes ...

@all-contributors add @ppalaga for code

shumonsharif:
@ppalaga @gastaldi @famod If you all are ok with it, I would like to propose giving @ppalaga maintainer status?

gastaldi:
@shumonsharif that is a great idea. Mind opening a PR changing https://github.com/quarkiverse/quarkiverse-devops/blob/main/quarkus-cxf.tf?

shumonsharif:
> @shumonsharif that is a great idea. Mind opening a PR changing https://github.com/quarkiverse/quarkiverse-devops/blob/main/quarkus-cxf.tf?

@ppalaga Are you ok with this?  Just wanted to get your permission before I put together the PR.  I believe @famod is [still on vacation](https://github.com/quarkiverse/quarkus-cxf/pull/467#issuecomment-1198403582), so I can go ahead if you provide a thumbs up.

famod:
I might be able to have a look at this today evening but don't hold your breath. 😉

ppalaga:
> If you all are ok with it, I would like to propose giving @ppalaga maintainer status?

That would be great, thanks for your trust!

.github/workflows/build.yml
```diff
@@ -36,5 +36,5 @@ jobs:
       run: ./mvnw -B formatter:validate install --file pom.xml
 
     - name: Build - Native
-      run: ./mvnw -B verify -Pnative -Dquarkus.native.container-build=true --file integration-tests/pom.xml
+      run: cd integration-tests && ../mvnw -B verify -Pnative -Dquarkus.native.container-build=true
```
famod, line 39: Nitpick: Why not `-f  integration-tests`?

integration-tests/basic/pom.xml
```diff
@@ -0,0 +1,138 @@
...
+                </exclusion>
+                <exclusion>
+                    <groupId>jakarta.xml.bind</groupId>
+                    <artifactId>jakarta.xml.bind-api</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.apache.geronimo.specs</groupId>
+                    <artifactId>geronimo-jta_1.1_spec</artifactId>
+                </exclusion>
+            </exclusions>
```
famod: Hum, aren't these exclusions already defined in the BOM?

famod (approved):
LGTM overall! 👍

.github/workflows/build.yml
ppalaga: `cd` tends to be safer for some buggy plugins (e.g. assuming that `Paths.get(".")` always returns `${basedir}`). Maybe we do not have any such here.

integration-tests/basic/pom.xml
ppalaga: No, `cxf-rt-transports-http-netty-server` is not in the BOM. It is a test dependency and it does not belong there, IMO. I'd like to remove it altogether in near future and replace it with a testcontainer.

ppalaga:
Rebased on top of the recent #487 and added some test refactoring commits.

.github/workflows/build.yml
famod: We probably don't but in the end I don't mind `cd`.

integration-tests/basic/pom.xml
famod: Are you sure? I see it here: https://github.com/quarkiverse/quarkus-cxf/pull/488/files#diff-954221b717ff3d93b2f171be5b7d728359dfa4ebee98e8e5691c0697682d247eR155

In fact, that very artifact was the main reason I added the BOM in the first place because I'm using it in my current project (for tests) and it's a maintenance burden to keep it in sync (without it in the BOM).

ppalaga: Ah, thanks for the evidence, then I have to correct myself: cxf-rt-transports-http-netty-server is not present in the *flattened* BOM (that gets effectively installed) because it is not a dependency of any of our CXF extensions. I'd prefer to keep that that way, because the BOM should be minimal IMO, covering just the end user runtime use cases.

If you find that particular artifact useful for testing, then I'd vote for putting it into a new quarkus-cxf-bom-test module much like in Quarkus and Camel Quarkus. WDYT?

famod: `quarkus-cxf-bom-test` makes sense!

IMO, the first release that contains this change (this PR) has to ship such a test bom as well, otherwise we'd be back to the pre-BOM state as far as that netty artifact is concerned, making it a bit of a breaking change.

cc @shumonsharif

ppalaga:
* Rebased on top of the current master
* Changed the description of this PR to match the content
* Ready for review

ppalaga:
Would anybody like to review?

shumonsharif:
LGTM, @ppalaga !

shumonsharif (approved):

PR #499, opened 2022-08-18
Refactor tests and QuarkusCxfProcessor

ppalaga:


ppalaga:
Would anybody like to review?

shumonsharif:
I can review in a couple hours.

shumonsharif (approved):
LGTM @ppalaga ... one minor comment is that CXF 3.5.3 depends on `woodstox-core` 6.2.8.  I'm not sure if switching that to 6.3.1 may have any unintended side effects?

ppalaga:
> LGTM @ppalaga ... one minor comment is that CXF 3.5.3 depends on `woodstox-core` 6.2.8. I'm not sure if switching that to 6.3.1 may have any unintended side effects?

Woodstox folks care for their backwards compatibility. Citing from their [release page](https://github.com/FasterXML/woodstox/wiki#releases)

> Woodstox follows [Apache versioning](https://apr.apache.org/versioning.html) convention, and similar to [Semantic Versioning](https://semver.org/) from external user perspective.

So going to a newer minor should not be a problem. Also, if our tests pass it should be fine. 
We can revert or report an issue to them if there are issues we do not see now.

PR #502, opened 2022-08-19
Test WS Security client 

### LINKED ISSUE #498: Test WS Security client
We have the quarkus-cxf-rt-ws-security extension, but no test covering it.

ppalaga:
Fix #498


shumonsharif (approved):
@ppalaga LGTM!  Just a minor comment - I didn't dig in too deeply, but it looked like there may be some duplication of code, and the README as well, so I was left wondering if these can somehow be abstracted and/or consolidated in some way?

ppalaga:
> there may be some duplication of code, and the README as well, so I was left wondering if these can somehow be abstracted and/or consolidated in some way?

+1, would you mind filing an issue, to make sure that I see what you see?

PR #509, opened 2022-08-23
Trigger release 1.5.0.CR1

ppalaga:


ppalaga:
Ups, it looks like the release job failed: https://github.com/quarkiverse/quarkus-cxf/runs/7968299448?check_suite_focus=true Anybody knows which GPG key is missing where? I do not have an access to the current project's secrets. @gastaldi maybe?

gastaldi:
@ppalaga you can't trigger the release from a forked repository. Do it from a branch in this repository and it should work

gastaldi:
Also please use `"` in the versions to avoid YAML issues while reading the versions

PR #510, opened 2022-08-23
Trigger release 1.5.0.CR1

ppalaga:
Attempt no. 2.

### DIFF
--- a/.github/project.yml
+++ b/.github/project.yml
@@ -1,5 +1,5 @@
 name: Quarkiverse Extension
 release:
-  current-version: 1.5.0.CR1
-  next-version: 1.5.0-SNAPSHOT
+  current-version: "1.5.0.CR1"
+  next-version: "1.5.0-SNAPSHOT"
 


ppalaga:
This worked 🎉

1.5.0.CR1 is available on Central: https://repo1.maven.org/maven2/io/quarkiverse/cxf/quarkus-cxf/1.5.0.CR1/

@famod @shumonsharif any feedback about this release would be welcome

famod:
@ppalaga I gave it a quick try in my current project (client only) and it's looking good so far. Great job!

shumonsharif:
@ppalaga Nice work, and great job indeed!  

My only minor suggestion is that the docs will need some updates, especially with regard to the introduction of the BOMs, though that doesn't necessarily need to coincide with the release.

ppalaga:
> the docs will need some updates, especially with regard to the recent introduction of the BOMs, though that doesn't necessarily need to coincide with the release.

Good idea. I filled https://github.com/quarkiverse/quarkus-cxf/issues/512 so that we do not forget.

PR #511, opened 2022-08-23
Move assertj out of bom-test, define common deps in integration-tests parent

famod:


integration-tests/pom.xml
```diff
@@ -50,5 +50,23 @@
...
     </dependencyManagement>
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.rest-assured</groupId>
+            <artifactId>rest-assured</artifactId>
```
famod, line 61: This is not actually used in one IT module: ws-security-server

But I don't see any real harm defining that very commong dep up here.

famod: I fact, we could pull up even more deps like testcontainers...

ppalaga (commented):
Thanks for taking the initiative, @famod !

I'd personally vote for minimal dependency set for each individual test, even in the test scope. I think it is generally better to be defensive and paranoid when testing. When adding an additional dependency, there is always a risk that it will interact with the rest in unexpected ways and thus cause test failures (or even false positive test passes) that we do not want to see. I do not have any specific failure scenario in mind. It is a sheer theoretical possibility. Call me methodological purity freak :D

famod:
I see your point in general.

OTOH, uniformity should also be our goal, i.e. we should use assertj wherever possible (and not a mix of junit assertions and assertj assertions), so I even see a reason to push assertj all the way up to to `quarkus-cxf-parent`.

What about meeting in the middle for now: I put rest-assured back to each individual module but leave the other two (which are used in all ITs) in the IT parent. WDYT?

Also /cc @shumonsharif for a third opinion.

ppalaga:
> What about meeting in the middle for now: I put rest-assured back to each individual module but leave the other two (which are used in all ITs) in the IT parent. WDYT?

+1, thanks!

famod:
Alright, I'll adjust this PR today evening.

ppalaga (approved):

shumonsharif:
@famod @ppalaga Sorry for being MIA, the day job has been keeping me very busy of late.

I think both your perspectives are very valid, and even though my personal preference has always been for uniformity, consistency and conciseness (If I were working on it, I would probably implement exactly as per @famod) - I'm glad there was some balance achieved by meeting in the middle.  I never quite thought of things as @ppalaga mentioned, and I'm sure there's good reason and possibly some hard experience backing it!

Thanks so much to you both for the flurry of recent activities - some rock solid work on the extension; I hope I'm able to more actively participate after a couple more weeks ...

PR #549, opened 2022-09-24
Cleanup keystore-generator module and elytron dependency

shumonsharif:
@ppalaga Our use case for keystores is fairly simple, and can be achieved using the keytool-maven-plugin fairly concisely, and without the dependency on elytron.  Please can you review this PR?


shumonsharif:
Aah, just noticed the failed native build ... will fix in a bit.

ppalaga (approved):
Wow, indeed a great simplification, thanks @shumonsharif !

Issue #551, opened 2022-09-29
Development mode reload is broken for exposed CXF services


jgrabowski:
I'm having CXF contract first service on Quarkus. While running in development mode each change generates failure saying:
Soap 1.1 endpoint already registered on address /smsprocessor
It looks like live reload cannot "release" already running webservice before applying changes. After restart it works fine until next code change.

shumonsharif:
This used to work fine before; linking issue below for reference:
https://github.com/quarkiverse/quarkus-cxf/issues/94

jgrabowski:
It was quite different. It wasn't' about unsatisfied dependencied but rather about conflict on context (endpoint already registered). I believe it's happening during Quarkus reload. I've enabled quarkus.live-reload.instrumentation=true and it doesn't crash anymore.

baptistejr:
Hello,
I have the same problem.
I think the problem came with this pr: https://github.com/quarkiverse/quarkus-cxf/pull/480

PR #559, opened 2022-10-04
Fix dev mode reload #551

shumonsharif:


shumonsharif:
@ppalaga Review would be welcome!

ppalaga:
@shumonsharif I looked through it and I wonder whether we could fully get rid of VertxDestinationRegistryFactory given that 

* We do not want to keep its static state between the reloads
* There is nothing really Vert.x specific in it
* We could just create a new DestinationRegistryImpl in CxfHandler()

I tried to implement that in https://github.com/quarkiverse/quarkus-cxf/pull/561 I am curious whether some client tests are going to fail. 

Plus I have added a dev mode test that we can keep regardless which solution will turn out to be better.

ppalaga:
https://github.com/quarkiverse/quarkus-cxf/pull/561 all tests passed, hence I'd vote for that fix because it is simpler.

ppalaga:
> how Camel is leveraging Quarkus CXF - I hope we don't break that integration? See [#480 (comment)](https://github.com/quarkiverse/quarkus-cxf/pull/480#issue-1326310689)

Haha, maybe I just shoot into my own foot? Thanks for the pointer! Let me check.

PR #561, opened 2022-10-05
Development mode reload is broken for exposed CXF services

### LINKED ISSUE #551: Development mode reload is broken for exposed CXF services

ppalaga:
Fix #551


shumonsharif (approved):
Super neat unit test!!  👍

ppalaga:
Let me check like this and check the damage in Camel Quarkus later because I have some subsequent commits ready now.

PR #655, opened 2022-12-14
Update release script

gastaldi:
- This removes the GPG file and bumps the actions versions


.github/workflows/release.yml
```diff
@@ -46,21 +50,33 @@ jobs:
...
-          ./mvnw -B release:perform -Darguments=-DperformRelease -Prelease -s maven-settings.xml
+          ./mvnw -B release:perform -Darguments=-DperformRelease -DperformRelease -Prelease
+        env:
+          MAVEN_USERNAME: ${{ secrets.OSSRH_USERNAME }}
+          MAVEN_PASSWORD: ${{ secrets.OSSRH_TOKEN }}
+
+      - name: Adjust tag for documentation changes
+        run: |
+          git checkout ${{steps.metadata.outputs.current-version}}
+          mvn -B clean install -DskipTests -DskipITs
```
ppalaga, line 71:
```suggestion
          ./mvnw -B clean install -DskipTests -DskipITs
```

.github/workflows/release.yml
```diff
@@ -46,21 +50,33 @@ jobs:
...
+          git commit -am "Update stable version for documentation"
+          # Move the tag after inclusion of documentation adjustments
+          git tag -f ${{steps.metadata.outputs.current-version}}
+          # Go back to base ref
+          git checkout ${{github.base_ref}}
```
ppalaga, line 76: What is the problem we are solving here?

gastaldi: This updates the attributes.adoc in the docs module with the latest available version

ppalaga: > This updates the attributes.adoc in the docs module with the latest available version

Does not this happen above around [here](https://github.com/quarkiverse/quarkus-cxf/blob/main/.github/workflows/release.yml#L54-L57)?

gastaldi: No, the released version in the `attributes.adoc` is not reflected in the main branch. This step sets the released version in that file

gastaldi: Actually, let me rectify that: it updates the main branch but it doesn't update the tag, just tested

ppalaga: So the new `Adjust tag for documentation changes` block does the same as the existing if block plus it moves the tag.

What I do not like about this is the duplicity and the fact that we move the tag after we have performed `release:perform` using the old tag.
I wonder whether we could not remove the new `Adjust tag for documentation changes` block and just add the tag reset `git tag -f ${{steps.metadata.outputs.current-version}}` at the end of the old if block?

gastaldi: AFAIK that won't work because in the `if` block the `[maven-release-plugin] prepare for next development iteration` commit already bumped to the next version, so you'll need to build without that commit and then rewrite the tag, which is what this does

.github/workflows/release.yml
```diff
@@ -46,21 +50,33 @@ jobs:
...
+          git checkout ${{steps.metadata.outputs.current-version}}
+          ./mvnw -B clean install -DskipTests -DskipITs
+          git commit -am "Update stable version for documentation"
+          # Move the tag after inclusion of documentation adjustments
+          git tag -f ${{steps.metadata.outputs.current-version}}
+          # Go back to base ref
+          git checkout ${{github.base_ref}}
 
       - name: Push changes to ${{github.base_ref}}
         uses: ad-m/github-push-action@v0.6.0
```
gastaldi, line 79: Let me change this too, do not merge yet

.github/workflows/release.yml
ppalaga: >  the [maven-release-plugin] prepare for next development iteration commit already bumped to the next version

Indeed, thanks explaining the obvious :) As long as I'll see some potential for improvements, I'll do it in a followup.

ppalaga (approved):

famod:
I haven't followed his closely but I'm going to merge this today (because we need to release #656).

Is this safe to backport to the 1.5 branch?

famod:
@gastaldi I must admit that it's my first time doing a quarkiverse extension release, so I might have done something wrong, but the release job failed with:
```
Error: Input required and not supplied: gpg-private-key
```
https://github.com/quarkiverse/quarkus-cxf/actions/runs/3730841563/jobs/6328362925

Can you point me in the right direction, please?

ppalaga:
@famod you need to push your trigger-release-1.7.2 branch to https://github.com/quarkiverse/quarkus-cxf instead of to your fork. 

Now that the release triggering commit is already merged in main, you have two options:

A. remove the commit from main, resend the PR unchanged and merge it again (I would choose this option)

B. Do some unimportant change (comment, add quotes, or similar) in `.github/project.yml` and send a new PR.

gastaldi:
Yes, as @ppalaga said. More information: https://github.com/quarkiverse/quarkiverse/wiki/Release#important

famod:
PS: Artifacts were deployed/uploaded already, but the release-commits were not pushed (because of the above error).

ppalaga:
Yeah, changes in release scripts usually require a couple of releases to become bug free :). Feel free to fix the release script @famod 

The missing commits and tag is a real issue. I'd vote for doing release:prepare locally to fix that. Would you like to do that @famod ?

famod:
> The missing commits and tag is a real issue. I'd vote for doing release:prepare locally to fix that. Would you like to do that @famod ?

You mean `./mvnw -B release:prepare -Prelease -DreleaseVersion=1.7.2 -DdevelopmentVersion=1.8.0-SNAPSHOT` and then push the commits?
I could do that.

gastaldi:
There is a small issue in the docs module, I fixed here: https://github.com/quarkiverse/quarkus-cxf/pull/661

ppalaga:
> You mean ./mvnw -B release:prepare -Prelease -DreleaseVersion=1.7.2 -DdevelopmentVersion=1.8.0-SNAPSHOT and then push the commits?

Yes, that should be it. Please make sure that the tag was also pushed.

famod:
Should be ok now.

FTR, I ran:
- `mvn -B release:prepare -Prelease -DreleaseVersion=1.7.2 -DdevelopmentVersion=1.8.0-SNAPSHOT -Darguments='-DskipTests -Dgpg.skip'`
- `git push upstream main`
- `git push upstream 1.7.2`

PR #664, opened 2022-12-20
Trigger release 1.5.10

famod:


famod:
Hum, @gastaldi can it be that we now have a conceptual problem with when `attributes.adoc` is changed?

https://github.com/quarkiverse/quarkus-cxf/actions/runs/3736247890/jobs/6340349003

The build fails because it detects a changed `attributes.adoc`, but a change is expected when touching `project.yml`...

gastaldi:
Hmm, the build.yml file is a bit different from what we recommend for Quarkiverse builds, maybe adapt to something similar?
  https://github.com/quarkusio/quarkus/blob/main/independent-projects/tools/base-codestarts/src/main/resources/codestarts/quarkus-extension/code/quarkiverse/java/.github/workflows/build.yml

famod:
@ppalaga you seem to have added the respective build step: https://github.com/quarkiverse/quarkus-cxf/commit/86ea22dab2757cde5a59f5645a1e6aea4c10ea25#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R42-R45
WDYT?

We could either remove the step, make it ignore attributes.adoc or don't run the attributes.adoc generation in normal build.
All options have their cons.

ppalaga:
I was able to release 1.5.10 after pushing some additional tweaks to release.yaml. Let me port the same changes to main

famod:
Ok, nice. But is it just me or are the tags missing for 1.7.3 and 1.5.10?

ppalaga:
> are the tags missing for 1.7.3 and 1.5.10?

Yes, I also noticed that the tag was missing for 1.5.10. I have pushed it manually and added a step in the release workflow to push it. 

I did not intend to release 1.7.3. I was just testing something and pushed the trigger-release-1.7.3 branch, but then removed it. 

Do we need 1.7.3 now?

famod:
I don't think we need one now, but maybe after #669?

PR #669, opened 2022-12-23
Fix #668 External DTD failure for uber-jar

### LINKED ISSUE #668: External DTD problem
After added in my pom the quarkus-cxf dependency, I receive a Fatal Error when building with quarkus-maven-plugin:
`[Fatal Error] :20:70: External DTD: Failed to read external DTD 'properties.dtd', because 'http' access is not allowed due to restriction set by the accessExternalDTD property.`

I tried to set the JVM arg `-Djavax.xml.accessExternalDTD=all` or `-Djavax.xml.accessExternalDTD=http` and `-Djavax.xml.accessExternalSchema=all` but nothing solved the issue.

I'm using quarkus-cxf version 1.7.2 and quarkus platform version 2.15.0.Final.

I noticed that the 'properties.dtd' is the one referenced in the wsdl.plugin.xml file inside cxf-rt-bindings-soap-3.5.5 jar (https://github.com/apache/cxf/blob/main/rt/bindings/soap/src/main/resources/META-INF/wsdl.plugin.xml)

and I also receive a warning after the fatal error:
`[WARNING] [io.quarkiverse.cxf.deployment.QuarkusCxfProcessor] can not merge wsdl.plugin.xml`

I'm building an uber-jar (don't know if it is relevant)

How can I solve the issue?
Thanks

shumonsharif:


extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
```diff
@@ -247,6 +247,8 @@ void buildXmlResources(BuildProducer<NativeImageResourceBuildItem> resources,
                         new GeneratedResourceBuildItem("META-INF/wsdl.plugin.xml", os.toByteArray()));
             }
 
+            System.clearProperty("javax.xml.accessExternalDTD");
```
famod, line 250: I think it would be "safer" to read the initial value before setting `"file, http"` and re-set that initial value instead of clearing the entire property?

shumonsharif: Good point @famod, will address this shortly!!

ppalaga:
@shumonsharif would you mind adding a test that fails before this change and succeeds after it? Otherwise it is hard to make sure that we won't break it in the future.

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/QuarkusCxfProcessor.java
```diff
@@ -211,8 +211,10 @@ void buildXmlResources(BuildProducer<NativeImageResourceBuildItem> resources,
         try (ByteArrayOutputStream os = new ByteArrayOutputStream()) {
             XPath xpath = XPathFactory.newInstance().newXPath();
             DocumentBuilderFactory documentBuilderFactory = DocumentBuilderFactory.newInstance();
-            documentBuilderFactory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
-            documentBuilderFactory.setAttribute(XMLConstants.ACCESS_EXTERNAL_SCHEMA, "");
+            documentBuilderFactory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
+            String originalDTDProperty = System.getProperty("javax.xml.accessExternalDTD");
+            System.setProperty("javax.xml.accessExternalDTD", "file, http");
```
ppalaga, line 216: I wonder whether we really need to set the system property for this to work? Wouldn't `documentBuilderFactory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "file,http");` have the same effect?

I am asking because setting the system property might lead to race conditions: the BuildStep methods are run in parallel and there might exist other concurrent BuildStep method setting the same property.

shumonsharif: Excellent point @ppalaga !  `documentBuilderFactory.setAttribute` should work fine, let me address this.

shumonsharif:
> @shumonsharif would you mind adding a test that fails before this change and succeeds after it? Otherwise it is hard to make sure that we won't break it in the future.

@ppalaga The only way I can think of to test the `@BuildStep` would be to execute a uber jar build and look through the console output.  Is this an agreeable approach?  I don't believe we have any tests for a `@BuildStep` currently, right?  Suggestions welcome in case you can think of a better approach!

ppalaga:
> the build appears to be failing in a seemingly unrelated unit test: io.quarkiverse.cxf.deployment.test.ServerDevModeTest

I could not reproduce it locally, so I just rerun the failing tests.

extensions/core/deployment/src/test/java/io/quarkiverse/cxf/deployment/test/UberJarMergedResourcesTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @RegisterExtension
+    static final QuarkusProdModeTest runner = new QuarkusProdModeTest()
+            .withApplicationRoot((jar) -> jar
+                    .addClass(UberJarMain.class))
+            .setApplicationName("uber-jar")
+            .setApplicationVersion("0.1-SNAPSHOT")
+            .setRun(true)
+            .setExpectExit(true)
+            .overrideConfigKey("quarkus.package.type", "uber-jar")
```
ppalaga, line 26: I suspect this is not doing what we'd hope for. These `@RegisterExtension` tests IIRC always work with an in-memory class-loader and I think `.overrideConfigKey("quarkus.package.type", "uber-jar")` has no effect. This can be proven by a simple experiment: when all changes in this PR except the test are reverted, the test should fail, but in reality it does not (I checked). I think we need a proper integration test. 

For reading the log from an integration test, you need two things:

1. `quarkus.log.file.enable = true` in [application properties](https://github.com/apache/camel-quarkus/blob/main/integration-test-groups/foundation/direct/src/main/resources/application.properties) 
2. A piece of Awaitility code repeatedly reading the whole log file: https://github.com/apache/camel-quarkus/blob/main/integration-test-groups/foundation/direct/src/test/java/org/apache/camel/quarkus/component/direct/it/DirectTest.java#L63-L66

shumonsharif: @ppalaga Reverting the PR (except for the unit test) failed the test for me when I had tried it locally?  Not sure why you're seeing something different - but I'll look into it later this week and keep you posted.

ppalaga: Yes, thanks for explaining, indeed it is failing as expected with those log messages modified.

ppalaga (approved):
Could you please squash your commits and rebase on top of the current main?

shumonsharif:
@ppalaga cc: @famod I just squashed my commits and rebased on top of current main.  However, the build is still oddly failing in that same `ServerDevModeTest` unit test.  I'm unable to recreate the build failure locally, it completes successfully on my laptop.  Any thoughts or suggestions on how to troubleshoot this would be welcome.

famod:
Without touching anything, it doesn't fail on my box either (WSL2 Ubuntu 22.04).

**But** I can make it fail via `-Dsurefire.runOrder=reversealphabetical` so it's a test ordering / test interferrence issue.
I'm having a closer look...

famod:
Passes:
```
 mvn clean install -f extensions/core/deployment/ -Dsurefire.runOrder=alphabetical -Dtest=UberJarMergedResourcesTest,ServerDevModeTest
```
Fails:
```
 mvn clean install -f extensions/core/deployment/ -Dsurefire.runOrder=reversealphabetical -Dtest=UberJarMergedResourcesTest,ServerDevModeTest
```

famod:
`RestAssured` in `ServerDevModeTest` somehow picks up the port of the previous prod mode test (`UberJarMergedResourcesTest`) which is 8081 but it should use 8080 instead.
Digging further...

famod:
I think this is a bug in Quarkus: It _seems_ that the `@QuarkusMain` class/method terminates so quickly after [`start()` in `QuarkusProdModeTest.beforeAll()`](https://github.com/quarkusio/quarkus/blob/2.15.2.Final/test-framework/junit5-internal/src/main/java/io/quarkus/test/QuarkusProdModeTest.java#L491) that [`.beforeEach()` is not seeing the process anymore and is starting it again](https://github.com/quarkusio/quarkus/blob/2.15.2.Final/test-framework/junit5-internal/src/main/java/io/quarkus/test/QuarkusProdModeTest.java#L757).
This is problematic because in `start()` there is also a [RestAssured setup call](https://github.com/quarkusio/quarkus/blob/2.15.2.Final/test-framework/junit5-internal/src/main/java/io/quarkus/test/QuarkusProdModeTest.java#L623) and so you have two setup calls and the original port before that test is lost.

I'll try to work around that and will at least report it upstream (or maybe even send a PR).

famod:
Green! 🎉 

@shumonsharif @ppalaga I'll go ahead and merge this. Please do raise concerns via separate issues or comments in case you disagree with my workaround.
If I find some time I'll release 1.7.3 on sunday evening.

shumonsharif:
> Green! 🎉
> 
> @shumonsharif @ppalaga I'll go ahead and merge this. Please do raise concerns via separate issues or comments in case you disagree with my workaround. If I find some time I'll release 1.7.3 on sunday evening.

Nice work Falko, and thank you for helping out with this.  The updates look good to me!

ppalaga:
Great troubleshooting, thanks a lot @famod !

PR #671, opened 2022-12-25
Fix #670: Address security findings, migrate LGTM to GitHub Code Scanning

### LINKED ISSUE #670: Address security findings, migrate from LGTM to GitHub Code Scanning
Both LGTM and GitHub Code Scanning are currently not available in the repository.  

PR #649 to enable CodeQL failed due to [1 Critical and 3 High security findings](https://github.com/quarkiverse/quarkus-cxf/pull/649/checks?check_run_id=9978525438) in the repository.  It seems like LGTM missed to identify these findings.

We need to address the critical and high findings before we can enable CodeQL / GitHub Code Scanning.

shumonsharif:
Fix #670


extensions/core/runtime/src/main/java/io/quarkiverse/cxf/transport/VertxHttpServletRequest.java
```diff
@@ -389,7 +385,12 @@ public Enumeration<String> getHeaders(String name) {
     public int getIntHeader(String name) {
         LOG.trace("getIntHeader");
         String v = getHeader(name);
-        return v == null ? -1 : Integer.parseInt(v);
+        try {
+            return v != null ? Integer.parseInt(v) : 0;
+        } catch (NumberFormatException nfe) {
+            LOG.error(nfe);
+        }
+        return 0;
```
ppalaga, line 393: This change breaks the contract defined in javax.servlet.http.HttpServletRequest.getIntHeader(String). The old implementation was correct IMO. What is GH Code scanning complaining about?

ppalaga (commented):
Thanks for taking care @shumonsharif ! One comment inline.

extensions/core/runtime/src/main/java/io/quarkiverse/cxf/transport/VertxHttpServletRequest.java
shumonsharif: Thanks for reviewing @ppalaga .  I wasn't completely sure on how to handle this one either, happy to revert it back to the original state -- here's the finding:  https://github.com/quarkiverse/quarkus-cxf/security/code-scanning/60

And the full list of findings is here in case you come across anything else:
https://github.com/quarkiverse/quarkus-cxf/security/code-scanning?query=pr%3A649+tool%3ACodeQL+is%3Aopen+

Please feel free to give a thumbs up if we should revert to the original code!

shumonsharif: I took a closer look at `javax.servlet.http.HttpServletRequest.getIntHeader(String)` and you are indeed correct, the change was breaking the contract.  I just reverted that to the original state.  Thanks again for the review!

ppalaga:
Thanks @shumonsharif !

PR #696, opened 2023-01-24
CxfHandler.loadClass() should not fail silently 

### LINKED ISSUE #625: CxfHandler.loadClass() should not fail silently
It currently just warns. It should throw an exception so that users do not need to struggle with typos in their application.properties

```
    private static Class<?> loadClass(String className) {
        try {
            return Thread.currentThread().getContextClassLoader().loadClass(className);
        } catch (ClassNotFoundException e) {
            // silent fail
        }
        try {
            return Class.forName(className);
        } catch (ClassNotFoundException e) {
            LOGGER.warn("failed to load class " + className);
            return null;
        }
    }
```

ppalaga:
Fix #625


famod (approved):
LGTM!

Issue #605, opened 2022-11-21
Upgrade to Quarkus 3.0/CXF 4.0/Jakarta 10


ppalaga:
Quarkus migration tips: https://github.com/quarkiverse/quarkiverse/wiki/Migrating-to-Quarkus-3.x
Blocked by CXF: https://issues.apache.org/jira/browse/CXF-8371

shumonsharif:
@ppalaga It may be possible to get started with development on this using the [CXF 4.0.0-SNAPSHOT](https://repository.apache.org/content/groups/snapshots/org/apache/cxf/cxf-bom/4.0.0-SNAPSHOT/) releases with the repositories included at the bottom of [this page](https://cxf.apache.org/docs/using-cxf-with-maven.html).

I ran a very crude check - all our `quarkus-cxf` dependencies resolved, and got the same output between the current 3.5.3 and the 4.0.0-SNAPSHOT versions:

```mvn dependency:tree | grep cxf | wc -l```

ppalaga:
Thanks for the tips, @shumonsharif !

I think I'll only have time to work on this after Dec 14th. In case you or anybody else would like to set up a branch/PR it would be welcome.

shumonsharif:
@ppalaga I just pushed https://github.com/quarkiverse/quarkus-cxf/tree/quarkus-3.x

Everything works fine except some of the native test executions.  I suspect the issues are unrelated to the extension; I will dig in as time permits, but feel free to do so as well!

shumonsharif:
Except for the couple native tests relating to MTOM, everything else should be fixed and working now.  For now, I have temporarily disabled `MtomIT` and `MtomAwtIT` unit tests, to see if the build passes.

@ppalaga @famod Please feel free to review as your time permits!

shumonsharif:
Looks like container based native builds are failing.  Without the below, the build on my laptop is passing fine ... will dig in again later!

-Dquarkus.native.container-build=true

shumonsharif:
Issues have been resolved with the native test failures.  There's an issue with running the mtom-awt native test on Mac OS, so it's [conditional on the OS](https://github.com/quarkiverse/quarkus-cxf/compare/quarkus-3.x#diff-707635b0580c7d741a6ec8571ea9882ea2a8de18ec765d8a4fde56c7afcbcaffR9-R13) for now.  Please take a look if you get a chance.

https://github.com/quarkiverse/quarkus-cxf/compare/quarkus-3.x

PR #708, opened 2023-02-12
Upgrade to Quarkus 3.0.0.Alpha4 and Jakarta 9

ppalaga:
Fix https://github.com/quarkiverse/quarkus-cxf/issues/605

Some part of this PR come from https://github.com/quarkiverse/quarkus-cxf/tree/quarkus-3.x by @shumonsharif


ppalaga:
5131a97: Fixed the CodeQL complaints

famod:
What's our policy w.r.t to Quarkus 2? Many apps will take a few months before jumping on Quarkus 3.
I have mixed feelings about `main` using an alpha version...?

/cc also @shumonsharif

ppalaga:
> What's our policy w.r.t to Quarkus 2? Many apps will take a few months before jumping on Quarkus 3. 

We have a well maintained 1.5 branch on top of well maintained Quarkus 2.13. I'd say that's the safest bet for Quarkus 2 users.

We could still create a new 1.x branch from the commit just before the upgrade to Quarkus 3 and release with Quarkus 1.16.

> I have mixed feelings about `main` using an alpha version...?

AFAIK, there are no plans for Quarkus 2.17 at this point. The next should be 3.0.0.Final (after some more Alphas/Betas). Quarkus 3.0.0 being such a big change, I thing it is good to concentrate on that one and test all incoming changes against it. I'd also release 2.0.0.Alpha1 of QCXF to be able to test in Camel Quarkus.

Issue #654, opened 2022-12-14
Dependabot is not picking up CXF updates


famod:
3.5.5 was released recently, 3.5.4 already 2 months ago, yet we are still on 3.5.3

_Edit: Update to 3.5.5 was done manually, so the above is not the case anymore. It's very likely though that the same will happen when 3.5.6 comes out._

https://mvnrepository.com/artifact/org.apache.cxf/cxf-core

Bot log:
```
updater | INFO <job_548992836> Checking if org.apache.cxf:cxf-core 3.5.3 needs updating
  proxy | 2022/12/14 11:18:47 [702] GET https://repo.maven.apache.org:443/maven2/io/quarkiverse/quarkiverse-parent/11/quarkiverse-parent-11.pom
  proxy | 2022/12/14 11:18:47 [702] 200 https://repo.maven.apache.org:443/maven2/io/quarkiverse/quarkiverse-parent/11/quarkiverse-parent-11.pom
  proxy | 2022/12/14 11:18:47 [704] GET https://repo.maven.apache.org:443/maven2/org/jboss/jboss-parent/39/jboss-parent-39.pom
  proxy | 2022/12/14 11:18:47 [704] 200 https://repo.maven.apache.org:443/maven2/org/jboss/jboss-parent/39/jboss-parent-39.pom
  proxy | 2022/12/14 11:18:47 [706] GET https://repository.jboss.org:443/nexus/content/groups/public/org/apache/cxf/cxf-core/maven-metadata.xml
  proxy | 2022/12/14 11:18:47 [706] 200 https://repository.jboss.org:443/nexus/content/groups/public/org/apache/cxf/cxf-core/maven-metadata.xml
  proxy | 2022/12/14 11:18:47 [708] HEAD https://repository.jboss.org:443/nexus/content/groups/public/org/apache/cxf/cxf-core/3.5.2-jbossorg-4/cxf-core-3.5.2-jbossorg-4.jar
  proxy | 2022/12/14 11:18:47 [708] 200 https://repository.jboss.org:443/nexus/content/groups/public/org/apache/cxf/cxf-core/3.5.2-jbossorg-4/cxf-core-3.5.2-jbossorg-4.jar
updater | INFO <job_548992836> Latest version is 3.5.2-jbossorg-4
updater | INFO <job_548992836> No update needed for org.apache.cxf:cxf-core 3.5.3
```
[...]

PR #723, opened 2023-02-27
Switch to Renovate from Dependabot for dependency updates (#654)

shumonsharif:
@famod @ppalaga Kindly share your thoughts on this draft PR?  If we decide to proceed on this, we will need to approve the https://github.com/quarkiverse/quarkiverse-devops/pull/125 PR before we handle this one.

Relevant issue for this PR is: https://github.com/quarkiverse/quarkus-cxf/issues/654

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,6 +13,9 @@ updates:
       - dependency-name: jakarta.xml.ws:jakarta.xml.ws-api
         versions:
           - ">= 3.0.0"
+    # Prevent PRs from opening except for security updates, given Renovate
+    # is used to manage regular dependency updates
+    open-pull-requests-limit: 0
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,11 +1,19 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:base", ":dependencyDashboard", ":rebaseStalePrs", ":enableVulnerabilityAlertsWithLabel('security')"
+    "config:base",
+    ":dependencyDashboard",
+    ":rebaseStalePrs",
+    ":enableVulnerabilityAlertsWithLabel('security')"
+  ],
+  "ignorePresets": [
+    ":prHourlyLimit2"
   ],
   "packageRules": [
     {
-      "matchPackagePatterns": ["io.quarkiverse.cxf"],
+      "matchPackagePatterns": [
+        "io.quarkiverse.cxf"
+      ],
       "enabled": false
     }
   ]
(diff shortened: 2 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -48,6 +48,7 @@ jobs:
         path: docs/target/site
 
     - name: Fail if there are uncommitted changes
+      if: ${{ github.actor != 'dependabot[bot]' && github.actor != 'renovate[bot]' }}
```
famod, line 51: I don't quite get this one. Can you elaborate please?

.github/dependabot.yml
```diff
@@ -13,6 +13,9 @@ updates:
       - dependency-name: jakarta.xml.ws:jakarta.xml.ws-api
         versions:
           - ">= 3.0.0"
+    # Prevent PRs from opening except for security updates, given Renovate
+    # is used to manage regular dependency updates
+    open-pull-requests-limit: 0
```
famod, line 18: I wonder if we should disable dependabot altogether? Otherwise we might have to manage exclusions in two ways?

.github/renovate.json
```diff
@@ -0,0 +1,20 @@
...
+    ":prHourlyLimit2"
+  ],
+  "packageRules": [
+    {
+      "matchPackagePatterns": [
+        "io.quarkiverse.cxf"
+      ],
+      "enabled": false
+    }
+  ]
```
famod, line 19: There is an ignore rule in the dependabot file: https://github.com/quarkiverse/quarkus-cxf/blob/main/.github/dependabot.yml#L13
Not sure it still applies after the switch to Quarkus 3?

famod (commented):
Thanks for moving this forward! I left a few comments.

.github/dependabot.yml
shumonsharif: The `open-pull-requests-limit: 0` setting will prevent dependabot from opening PRs for any regular version updates.  It will not, however, prevent security updates from coming through.   I'm not entirely sure whether we'll end up with two PRs for each security update (one from renovate and one from dependabot).  I haven't been able to find much in terms of documentation on this topic, other than at the link below:

https://docs.renovatebot.com/configuration-options/#vulnerabilityalerts
"Renovate can read GitHub's Vulnerability Alerts to customize its Pull Requests. For this to work, you must enable the [Dependency graph](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph#enabling-the-dependency-graph), and [Dependabot alerts](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-security-and-analysis-settings-for-your-repository)."

My thoughts are we keep both dependabot and renovate running for now, and make a call on disabling dependabot altogether once we see how things play out.  Worst case, we will end up with some duplicate PRs for security updates, but these should be very infrequent.

.github/renovate.json
shumonsharif: Yes, I don't believe the exclusion applies any longer given our Quarkus 3 / Jakarta updates.

Issue #714, opened 2023-02-16
Make wsdl2java part of Quarkus code generation phase


ppalaga:
Quarkus extensions can generate Java source code prior to compilation. It works like that for [gRPC](https://github.com/quarkusio/quarkus/blob/2d4fe75178cb3abda2af81ffe7cb24244e3f5508/extensions/grpc/codegen/src/main/java/io/quarkus/grpc/deployment/GrpcCodeGen.java) and other extensions. The objective here is to replace the usage of `cxf-codegen-plugin` with a piece of code inside our core extension that would wrap the `cxf-tools-wsdlto-*` CXF libraries.

scrocquesel:
I start implementing a simple codegen https://github.com/scrocquesel/quarkus-cxf/tree/gh-714. I first try calling wsdl1java directly but using `cxf-tools-wsdlto-*` seems easier. I just struggle with a jaxb dependency version mismatch.

wip at https://github.com/scrocquesel/quarkus-cxf/tree/gh-714

ppalaga:
Great work, thanks @scrocquesel! Could you perhaps send a draft pull request so that I can see the failures in the CI log without having to rebuild your branch locally?

scrocquesel:
> Great work, thanks @scrocquesel! Could you perhaps send a draft pull request so that I can see the failures in the CI log without having to rebuild your branch locally?

@ppalaga for now, I only change the it/client project to use the quarkus codegen plugin.

PR #731, opened 2023-02-27
codegen

scrocquesel:
early poc for #714 
We need to clarify which options should be supported


ppalaga:
Replaced by https://github.com/quarkiverse/quarkus-cxf/pull/751

PR #747, opened 2023-03-03
Handle @sync dependency updates

shumonsharif:
@ppalaga cc: @famod Please review this PR, I was hoping to address the issues outlined in https://github.com/quarkiverse/quarkus-cxf/pull/739#issuecomment-1451920919

I believe most CVEs would typically come through as either a minor or patch release ... but wasn't sure what makes the most sense in terms of how I have grouped the dependencies in the PR.  Please share your thoughts.

If we need to add specific dependency version ranges, that can also easily be done using:
https://docs.renovatebot.com/configuration-options/#allowedversions

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -11,8 +11,36 @@
   ],
   "packageRules": [
     {
+      "description": "Ignore major updates for matched packages (allow only minor and patch updates)",
       "matchPackagePatterns": [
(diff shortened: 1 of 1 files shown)


ppalaga (approved):
Thanks @shumonsharif, looks good! Let's merge this and see whether Renovate will behave as expected.

famod (approved):
Agreed

PR #755, opened 2023-03-12
Add WSDLs processed by wsdl2java to native image automatically

ppalaga:
There are two main changes here:

1. Change default source directory of wsdl2java from src/main/wsdl to src/main/resources so that the WSDLs are incuded in the application and so that constructs like wsdlLocation = "classpath:foo.wsdl" work
2. Add WSDLs processed by wsdl2java to native image automatically

The main motivation is that without these it is very cumbersome to get an embedded WSDL working with `@WebService( wsdlLocation = "foo.wsdl")` or `@WebServiceClient(wsdlLocation = "classpath:wsdl/foo.wsdl")`. 

I also deprecated `CxfBuildTimeConfig.wsdlPath` in favor of 2. and the stock Quarkus `quarkus.cxf.codegen.wsdl2java.includes/excludes`.

I'll release another Alpha shortly, so that I can integrate it in Camel Quarkus.

cc @scrocquesel @llowinge


scrocquesel:
That also gives opportunity to import wsdl from dependency

ppalaga:
> wsdl from dependency

I do not think this would work ATM. Would you please file an issue if you need that?

PR #778, opened 2023-03-23
CI & mvnd related improvements

ppalaga:


ppalaga:
Now it should be possible to build in parallel with `mvnd clean install -DskipTests -Dquarkus.build.skip`

I would not recommend running tests in parallel, because quarkus-maven-plugin:build is not able to be run in parallel - that's why `-Dquarkus.build.skip`

Issue #582, opened 2022-10-28
Unmarshalling Error: unexpected element when invoking a service with multiple parameters without explicit wrapper


ppalaga:
There is a disabled test reproducing this issue:

https://github.com/quarkiverse/quarkus-cxf/blob/226d033ffdde3785a3383bd6b565992d12e69711/integration-tests/mtom-awt/src/test/java/io/quarkiverse/cxf/it/ws/mtom/awt/server/MtomAwtTest.java#L27-L31

The stack trace: 

```
[INFO] Running io.quarkiverse.cxf.it.ws.mtom.awt.server.MtomAwtTest
[INFO] [stdout] 2022-10-28 17:44:40,183 INFO  [io.qua.cxf.dep.CxfEndpointImplementationProcessor] (build-16) Mapping a Vert.x handler for CXF to /soap/* as requested by [io.quarkiverse.cxf.it.ws.mtom.awt.server.ImageServiceImpl]
[INFO] [stdout] 2022-10-28 17:44:41,088 INFO  [org.apa.cxf.end.ServerImpl] (main) Setting the server's publish address to be /mtom-aws
[INFO] [stdout] 2022-10-28 17:44:41,095 INFO  [io.qua.cxf.tra.CxfHandler] (main) Web Service io.quarkiverse.cxf.it.ws.mtom.awt.server.ImageServiceImpl on /soap available.
[INFO] [stdout] 2022-10-28 17:44:41,225 INFO  [io.quarkus] (main) quarkus-cxf-integration-test-mtom-awt 1.6.0-SNAPSHOT on JVM (powered by Quarkus 2.13.3.Final) started in 2.874s. Listening on: http://localhost:8081
[INFO] [stdout] 2022-10-28 17:44:41,226 INFO  [io.quarkus] (main) Profile test activated. 
[INFO] [stdout] 2022-10-28 17:44:41,226 INFO  [io.quarkus] (main) Installed features: [awt, cdi, cxf, cxf-rt-features-logging, resteasy, smallrye-context-propagation, vertx]
[INFO] [stdout] 2022-10-28 17:44:42,710 INFO  [org.apa.cxf.ser.Ima.REQ_IN] (executor-thread-1) REQ_IN
[INFO] [stdout]     Address: http://localhost:8081/soap/mtom-aws/ImageService
[INFO] [stdout]     HttpMethod: POST
[...]

PR #780, opened 2023-03-26
WiP Unmarshalling Error: unexpected element when invoking a service with

ppalaga:
multiple parameters without explicit wrapper 

This is an attempt to f.i.x #582. While the `Unmarshalling Error: unexpected element` goes away, the fix unfortunately causes another issue with unexpected namespace declaration on elements that pops up in nearly all other tests. So this PR is perhaps only useful to show what does not work. Any ideas to f.i.x #582 are welcome!


ppalaga:
Replaced by https://github.com/quarkiverse/quarkus-cxf/pull/791

PR #796, opened 2023-04-03
WiP Default value / of quarkus.cxf.path makes RESTeasy and health endpoints inaccessible 

### LINKED ISSUE #795: Default value `/` of `quarkus.cxf.path` makes RESTeasy and health endpoints inaccessible
Originally reported in https://github.com/apache/camel-quarkus/issues/4192

We should check, whether we can register our Vert.x handler without using the wildcard path matcher `/*`.

ppalaga:
Fix #795

### DIFF
--- a/integration-tests/ws-trust/src/test/java/io/quarkiverse/cxf/it/ws/trust/CxfWsTrustTest.java
+++ b/integration-tests/ws-trust/src/test/java/io/quarkiverse/cxf/it/ws/trust/CxfWsTrustTest.java
@@ -83,7 +83,7 @@ public class CxfWsTrustTest {
             final QName serviceName = new QName("https://quarkiverse.github.io/quarkiverse-docs/quarkus-cxf/test/ws-trust",
                     "TrustHelloService");
             final URL wsdlURL = new URL(io.quarkiverse.cxf.test.QuarkusCxfClientTestUtil.getServerUrl()
-                    + "/jaxws-samples-wsse-policy-trust/TrustHelloService?wsdl");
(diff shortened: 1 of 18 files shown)


integration-tests/ws-trust/src/test/java/io/quarkiverse/cxf/it/ws/trust/CxfWsTrustTest.java
```diff
@@ -83,15 +83,15 @@ public void programmaticSts() throws Exception {
             final QName serviceName = new QName("https://quarkiverse.github.io/quarkiverse-docs/quarkus-cxf/test/ws-trust",
                     "TrustHelloService");
             final URL wsdlURL = new URL(io.quarkiverse.cxf.test.QuarkusCxfClientTestUtil.getServerUrl()
-                    + "/jaxws-samples-wsse-policy-trust/TrustHelloService?wsdl");
+                    + "/jaxws-samples-wsse-policy-trust?wsdl");
```
ppalaga, line 86: @ffang, @shumonsharif, @famod, any opinion, whether it is fine to generally stop supporting `/my-path/MyService?wsdl`? Note that `/my-path?wsdl` worked before this change. I actually suspect that `/my-path/AnyString?wsdl` worked before too, which sounds like a bug to me that we are hereby fixing. But I admit I have little knowledge about what is normal in the world of CXF and therefore I am asking.

ppalaga:
Let's merge this and revisit if needed.

ppalaga:
This unfortunately breaks Camel. I have to revert it for now.

Issue #797, opened 2023-04-04
Wrap `java2ws` in Quarkus build time tooling


ppalaga:
Initially, generating WSDL out of Java would be enough. 

We already support `wsdl2java` via [Wsdl2JavaCodeGen](https://github.com/quarkiverse/quarkus-cxf/blob/75698ee1afbd6fb350b1aa3a9e0e2e54540d5d86/extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/codegen/Wsdl2JavaCodeGen.java) `java2ws` should work in the same way. We need to add some config and a new CodeGen class wrapping `java2ws`.

JiriOndrusek:
hi @ppalaga , may I be assigned to this issue? I'd like to fix it.

ppalaga:
Thanks, done, @JiriOndrusek

JiriOndrusek:
@ppalaga I've spent some time investigation this issue and I feel that it is not possible (or at least beneficial for users). Let me explain:

- `java2ws` requires compiled java class to be able to generate *.wsdl from it
- my understanding of `quarkus-maven-plugin` is that it is called in `generate-code` of the maven lifecycle, therefore it doesn't have acceess to compiled classes
- if this functionality was happening after the compilation, what would be beneficial opposed of using `cxf-java2ws-plugin`?

I probably miss some relation to make it sense for me.

Does my concern make sense to you? From the doc of the `CodeGenProvider`

```
Service providers for this interface are triggered during generate-sources phase of build of Quarkus applications
```

JiriOndrusek:
Hi @ppalaga 
I'm facing 2 issues. 
• My initial idea was to use `CombinedIndexBuildItem` to find all instances of `@WebService`, filter them  with includes and excludes (using regexPattern). Once I add `CombinedIndexBuildItem` as an input parameter of my build step, I'm getting a Quarkus exception `The build step is not currently running` (during execution)
• If I run the `java2ws` tool multiple times, I'm getting the same issue (`The build step is not currently running`)

In case that combinedIndexBuildItem is not present in my build step and there is only 1 execution of `java2ws` tool, the execution is successful and wsdl is generated.

Do you have an idea, what could be wrong with my idea?

ppalaga:
Could you please share your code?

JiriOndrusek:
Sure.
Here is my branch https://github.com/JiriOndrusek/quarkus-cxf/tree/797-java2wsdl-works-once

[...]

PR #815, opened 2023-04-13
Support java2ws in Quarkus code generation phase #797

JiriOndrusek:
fixes https://github.com/quarkiverse/quarkus-cxf/issues/797

There are 4 small problems, marked by `todo` in the comment of the code:
* there is a workaround in QuarkusCapture, which is not correct (eventhough it works, but may show possible problems)
* I had to keep `BuildProducer<ReflectiveClassBuildItem>` in `Java2WsdlProcessor.java2wsdl` to force Qiarkus to execute his build step. If producer  is not present, Quarkus ignores step because there is no output from it. I was not  able to find a proper solution for such scenario (I found `ArtifactResultBuildItem` which should be used, but item didn't work for me)
* java2ws feature is disabled by defaul. wsdl2java is enabled by default -> should behave probably in the same way
* exclusion of `vaadin-development-mode-detector:<groupId>org.mvnpm.at.vaadin` should not be necessary, but I can not build the module locally without it.

### DIFF
--- a/docs/modules/ROOT/pages/includes/quarkus-cxf.adoc
+++ b/docs/modules/ROOT/pages/includes/quarkus-cxf.adoc
@@ -98,6 +98,115 @@ endif::add-copy-button-to-env-var[]
 |
 
 
+a|icon:lock[title=Fixed at build time] [[quarkus-cxf_quarkus.cxf.java2ws.enabled]]`link:#quarkus-cxf_quarkus.cxf.java2ws.enabled[quarkus.cxf.java2ws.enabled]`
+
+[.description]
+--
+If `true` `java2ws` wsdl generation is run whenever there are Java classes annotated with `jakarta.jws.WebService` selected via following properties; otherwise `java2ws` is not executed.
+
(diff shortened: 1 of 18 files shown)


JiriOndrusek:
@ppalaga FYI ^^^

JiriOndrusek:
@ppalaga  Should I make the tests more detailed, to compare content of generated wsdl to expected one?

JiriOndrusek:
As I see [Remove the **.wsdl default for quarkus.cxf.codegen.wsdl2java.includes](https://github.com/quarkiverse/quarkus-cxf/pull/816#issue-1667779861)
I used different approach. I kept the defailt include (`.*` in this case)`, but set the default value `false` int to `*.enabled` property.
I suppose, that I should use the same approach as wsdl2java, right?

ppalaga:
Thanks @JiriOndrusek! I amended some changes in https://github.com/quarkiverse/quarkus-cxf/pull/820 , thus closing this one.

JiriOndrusek:
Thanks!

PR #816, opened 2023-04-14
Remove the `**.wsdl` default for quarkus.cxf.codegen.wsdl2java.includes

ppalaga:
I am in favor of removing the quarkus.cxf.codegen.wsdl2java.includes default after getting feedback from @famod, @llowinge and @JiriOndrusek. 
@famod mentioned somewhere that the default would typically conflict with whatever named parameter set defined by the end user via quarkus.cxf.codegen.wsdl2java.foo-params.includes unless she narrows quarkus.cxf.codegen.wsdl2java.includes.
@JiriOndrusek mnetioned that the default may cause funny things under some circumstances once we have also java2wsdl.
So I hope there is a general agreement that wsdl2java should not be enabled without end user specifying an include.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -70,6 +70,12 @@
                 <groupId>org.apache.cxf</groupId>
                 <artifactId>cxf-rt-transports-http-hc5</artifactId>
                 <version>${cxf.version}</version>
+                <exclusions>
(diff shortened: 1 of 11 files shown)


llowinge:
That makes sense, thanks @ppalaga !

PR #841, opened 2023-04-23
Manage all AsciiDoc attributes in antora.yml

ppalaga:


ppalaga:
@gastaldi I find having all AsciiDoc attributes in antora.yaml a bit better because it that way I do not need to include attributes.adoc to every single doc page.

FYI, we also changed the setup for testing the site. Now it is based on [Antora TestContainer](https://github.com/quarkiverse/quarkus-cxf/blob/main/docs/src/test/java/io/quarkiverse/cxf/doc/it/AntoraTest.java). It feels much more reality close, easier to maintain and easier to use for visual checks of the Antora output than the previous setup with asciidoctor-maven-plugin.

gastaldi:
@ppalaga that's an interesting approach, I wonder if we can reuse some of that stuff to minimize the number of files to be maintained

ppalaga:
> I wonder if we can reuse some of that stuff to minimize the number of files to be maintained

If you ask me, I'd find it awesome if the code from https://github.com/quarkiverse/quarkus-cxf/blob/main/docs/src/test/java/io/quarkiverse/cxf/doc/it/AntoraTest.java could be packed into a quarkus extension that would serve the generated site on localhost:8080 with live reload included ;)

PR #845, opened 2023-04-24
Specify which parts of CXF are supported

ppalaga:
I tried to define which parts of CXF we support. @shumonsharif @famod your review would be highly appreciated.

### DIFF
--- a/docs/modules/ROOT/examples/soap-service/application.properties
+++ b/docs/modules/ROOT/examples/soap-service/application.properties
@@ -56,6 +56,8 @@ quarkus.cxf.client."greetingclient".client-endpoint-url=http://localhost:8081/so
 quarkus.cxf.client."greetingclient".service-interface=io.quarkiverse.cxf.it.server.GreetingClientWebService
 
 # java2ws
+# tag::generate-wsdl-from-java.adoc[]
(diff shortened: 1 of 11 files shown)


ppalaga:
cc @vkasala

shumonsharif:
@ppalaga This is great!  

A few minor thoughts; I'm not convinced myself that we need to call out the below explicitly as supported on the "Reference" page, so please feel free to disregard if you don't feel it'll fit into the overall context of what we're trying to capture (I'm assuming the intent is to capture all supported features and configurations in one page).

- We support most of the CXF [Annotations](https://cxf.apache.org/docs/annotations.html)
- We support custom CXF [Features](https://cxf.apache.org/docs/features.html) and [Interceptors](https://cxf.apache.org/docs/interceptors.html) - developers can plug in their own implementations.
- Don't know if it's possible to squeeze into the Reference page; perhaps a small info icon with a tooltip containing the explanations of why 'Experimental' extensions are marked as such.

ppalaga:
I am all for it! Would you like to send a PR? I have created https://github.com/quarkiverse/quarkus-cxf/issues/865 so that we do not forget.

Issue #504, opened 2022-08-22
Test quarkus-cxf-rt-features-metrics


ppalaga:
We should add a separate integration test for quarkus-cxf-rt-features-metrics-parent, much like we have for the other extensions.

ppalaga:
Anybody is using `quarkus-cxf-rt-features-metrics` with micrometer and could sketch how it could be tested?

ppalaga:
Great, thanks a lot @shumonsharif for sharing the steps and for being ready to add a test!
I guess the assignee field of this issue is the best semaphore for avoiding conflicts. Just assign to yourself once you start working on this. And the same holds for others: if anybody would like to try before @shumonsharif has time, assign to yourself or ask for assigning to you.

ppalaga:
Let me try this before releasing 2.2.0.

PR #915, opened 2023-06-10
initialize CodahaleMetricsProvider at runtime

knuspertante:
Native-Image reports, that CodahaleMetricsProvider should be initialized at runtime.

`Error: Class initialization of org.apache.cxf.metrics.codahale.CodahaleMetricsProvider failed. This error is reported at image build time because class org.apache.cxf.metrics.codahale.CodahaleMetricsProvider is registered for linking at image build time by command line Use the option --initialize-at-run-time=org.apache.cxf.metrics.codahale.CodahaleMetricsProvider to explicitly request delayed initialization of this class.
`


ppalaga:
@knuspertante thanks for the contribution! However it is hard to accept this as is without a test. Would you perhaps find any time to write one? There are some hints in https://github.com/quarkiverse/quarkus-cxf/issues/504 and you could also take some inspiration from other existing integration tests in https://github.com/quarkiverse/quarkus-cxf/tree/main/integration-tests

ppalaga:
Thanks again, for your contribution @knuspertante! I finally found some time to write the test and also polish the documentation: https://github.com/quarkiverse/quarkus-cxf/pull/932 Would you like to review?

Issue #916, opened 2023-06-10
Could not resolve artifact jakarta.xml.ws:jakarta.xml.ws-api::jar


DMaxter:
Hello!

I'm currently using Quarkus 2.14.0.Final (not yet upgraded to Quarkus 3) with gradle.
I need to interact with a SOAP API so I just added `implementation("io.quarkiverse.cxf:quarkus-cxf")` to the dependencies of my project. 
I followed the steps in [Your first SOAP Client on Quarkus](https://quarkiverse.github.io/quarkiverse-docs/quarkus-cxf/dev/user-guide/first-soap-client.html) and [Generate the Model classes from WSDL](https://quarkiverse.github.io/quarkiverse-docs/quarkus-cxf/dev/user-guide/generate-java-from-wsdl.html), but when trying to run Quarkus (without adding any code, just trying to obtain the generated classes) I got this exception.

```
Failed to start quarkus: java.lang.RuntimeException: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
        [error]: Build step io.quarkus.deployment.index.ApplicationArchiveBuildStep#build threw an exception: java.lang.RuntimeException: Could not resolve artifact jakarta.xml.ws:jakarta.xml.ws-api::jar among the runtime dependencies of the application
        at io.quarkus.deployment.index.ApplicationArchiveBuildStep.addIndexDependencyPaths(ApplicationArchiveBuildStep.java:188)
        at io.quarkus.deployment.index.ApplicationArchiveBuildStep.scanForOtherIndexes(ApplicationArchiveBuildStep.java:151)
        at io.quarkus.deployment.index.ApplicationArchiveBuildStep.build(ApplicationArchiveBuildStep.java:108)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:568)
        at io.quarkus.deployment.ExtensionLoader$3.execute(ExtensionLoader.java:909)
        at io.quarkus.builder.BuildContext.run(BuildContext.java:281)
        at org.jboss.threads.ContextHandler$1.runWith(ContextHandler.java:18)
[...]

Issue #860, opened 2023-05-02
io.quarkus.builder.ChainBuildException: Cycle detected


guitcastro:
Hello, I am trying to add the extension on a new project (ext version version 2.0.1 and quarkus 3.0.1-Final). The code is generated but the build fail with:

Execution failed for task ':infrastructure:quarkusAppPartsBuild'.
> There was a failure while executing work items
   > A failure occurred while executing io.quarkus.gradle.tasks.worker.BuildWorker
      > io.quarkus.builder.ChainBuildException: Cycle detected:
                           io.quarkiverse.cxf.deployment.CxfClientProcessor#collectClients produced class io.quarkiverse.cxf.deployment.ClientSeiBuildItem
                        to io.quarkiverse.cxf.deployment.QuarkusCxfProcessor#generateClasses produced class io.quarkus.arc.deployment.GeneratedBeanBuildItem
                        to io.quarkus.arc.deployment.BeanArchiveProcessor#build produced class io.quarkus.arc.deployment.BeanArchiveIndexBuildItem
                        to io.quarkus.arc.deployment.LookupConditionsProcessor#suppressConditionsGenerators produced class io.quarkus.arc.deployment.SuppressConditionGeneratorBuildItem
                        to io.quarkus.arc.deployment.ArcProcessor#initialize produced class io.quarkus.arc.deployment.ContextRegistrationPhaseBuildItem
                        to io.quarkus.arc.deployment.ArcProcessor#registerBeans produced class io.quarkus.arc.deployment.BeanRegistrationPhaseBuildItem
                        to io.quarkus.vertx.deployment.VertxProcessor#collectEventConsumers produced class io.quarkus.vertx.deployment.EventConsumerBusinessMethodItem
                        to io.quarkus.vertx.deployment.VertxProcessor#build produced class io.quarkus.vertx.deployment.VertxBuildItem
                        to io.quarkus.reactive.pg.client.deployment.ReactivePgClientProcessor#build produced class io.quarkus.deployment.builditem.ExtensionSslNativeSupportBuildItem
                        to io.quarkus.deployment.steps.NativeImageConfigBuildStep#build produced class io.quarkus.deployment.builditem.nativeimage.RuntimeInitializedClassBuildItem
                        to io.quarkiverse.cxf.deployment.CxfClientProcessor#collectClients

```

ghermaniov:
Hi,

I am having the same issue. This is my stack trace:

`Cycle detected:
[...]

Issue #580, opened 2022-10-28
Clients with a class with postponed initialization in their method signatures cannot be compiled to native


ppalaga:
Client proxies for the service interfaces (SEI) are generated at runtime by GraalVM native-image. That generated code loads all classes present in all method signatures of the SEI. Hence if we have a SEI like the following

```
@WebService
interface Foo {
  Bar doSomething(Baz baz);
}
```

then both `Bar` and `Baz` classes get loaded via `Class.forName("acme.Bar")` and `Class.forName("acme.Baz")` in the Proxy code generated by GraalVM native-image (see the listing [below](https://github.com/quarkiverse/quarkus-cxf/issues/580#issuecomment-1296352415)).

In Quarkus all classes are initialized at build time except for those specific classes for which runtime init is requested via `RuntimeInitializedClassBuildItem`. Hence the generated Proxy class is by default also initialized at build time. As a consequence of that all classes mentioned in method signatures of client SEIs are also initialized at build time.

If some if those classes needs to be initialized at runtime (such as `java.awt.Image`) and is registered via `RuntimeInitializedClassBuildItem`, then the native compilation fails with `Error: Classes that should be initialized at run time got initialized during image building`

There is a reproducer in https://github.com/quarkiverse/quarkus-cxf/pull/579

The stack trace: 

```
[INFO] [stdout] ========================================================================================================================
[INFO] [stdout] GraalVM Native Image: Generating 'quarkus-cxf-integration-test-mtom-awt-1.6.0-SNAPSHOT-runner' (executable)...
[INFO] [stdout] ========================================================================================================================
[INFO] [stdout] [1/7] Initializing...                                                                                    (7.8s @ 0.17GB)
[INFO] [stdout]  Version info: 'GraalVM 22.2.0 Java 17 CE'
[INFO] [stdout]  Java version info: '17.0.4+8-jvmci-22.2-b06'
[INFO] [stdout]  C compiler: gcc (redhat, x86_64, 8.5.0)
[INFO] [stdout]  Garbage collector: Serial GC
[INFO] [stdout]  6 user-specific feature(s)
[INFO] [stdout]  - io.quarkus.awt.runtime.graal.AwtFeature
[INFO] [stdout]  - io.quarkus.awt.runtime.graal.DarwinAwtFeature
[INFO] [stdout]  - io.quarkus.runner.Feature: Auto-generated class by Quarkus from the existing extensions
[...]

PR #937, opened 2023-07-06
Replace proxy runtime init code with "-H:+UseNewExperimentalClassInit…

shumonsharif:
…ialization"

@ppalaga I'm wondering if the -H:+UseNewExperimentalClassInitialization flag is a viable alternative to the Proxy runtime initialization workarounds we put in place.  This approach also resolves all the cyclic issues identified in #860 and #916 .

EDIT:  Looking closer through all the previous notes on Proxy runtime init, I just stumbled upon https://github.com/quarkiverse/quarkus-cxf/issues/580#issuecomment-1481237043 where https://github.com/oracle/graal/pull/4684 was already referenced.  I guess I could've saved a couple hours of digging if I had read all of the notes carefully!!

I'm curious why you didn't go with this option, as it seems to be only a matter of time before this becomes the new default in Graal?

### DIFF
--- a/docs/modules/ROOT/examples/calculator-client/application.properties
+++ b/docs/modules/ROOT/examples/calculator-client/application.properties
@@ -42,6 +42,6 @@ quarkus.cxf.client.clientWithRuntimeInitializedPayload.service-interface = io.qu
 quarkus.cxf.client.clientWithRuntimeInitializedPayload.endpoint-namespace = http://www.jboss.org/eap/quickstarts/wscalculator/Calculator
 quarkus.cxf.client.clientWithRuntimeInitializedPayload.endpoint-name = CalculatorService
 quarkus.cxf.client.clientWithRuntimeInitializedPayload.native.runtime-initialized = true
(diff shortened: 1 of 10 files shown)


dufoli:
What is the stage of this graal feature ?

shumonsharif:
Hi @dufoli I'm not quite sure which stage this feature is in, or any specific timelines around it.  Pretty much the only info I have is in the referenced graal issue.  

It seems like a far simpler solution (even though the flag is experimental) to address the very unique edge case of proxy runtime initialization.  I feel like we may be able to live with this flag until it becomes the Graal default - only because the list of extensions quarkus-cxf is currently incompatible with due to the cycle issues, seems to keep growing.

andre-s-sousa:
Hey guys, any idea when this may get merged and released?
Sadly the loop bug is causing quite some issues that I have no solution for

dufoli:
@ppalaga what do you think about it ?

ppalaga:
I finally found time to start looking into this. Sorry for the delay.

ppalaga:
I have commented in https://github.com/quarkiverse/quarkus-cxf/issues/860#issuecomment-1679440730 and in accordance with that I am for closing the current PR in favor of https://github.com/quarkiverse/quarkus-cxf/pull/963

PR #938, opened 2023-07-07
Simplify the doc title

gastaldi:
- According to https://github.com/quarkiverse/quarkiverse-devops/pull/164#discussion_r1255308601

### DIFF
--- a/docs/antora.yml
+++ b/docs/antora.yml
@@ -19,7 +19,7 @@ asciidoc:
 
     # External URLs
     link-quarkus-code-generator: code.quarkus.io
-    quarkus-cxf-project-name: CXF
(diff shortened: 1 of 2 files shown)


docs/antora.yml
```diff
@@ -19,7 +19,7 @@ asciidoc:
 
     # External URLs
     link-quarkus-code-generator: code.quarkus.io
-    quarkus-cxf-project-name: CXF Extensions for Quarkus
+    quarkus-cxf-project-name: CXF
```
ppalaga, line 22: This should stay as is because it is the full project name that we use throughout the docs.

ppalaga:
Thanks @gastaldi, the CI failure is unrealted.

PR #956, opened 2023-08-09
Fix asciidoc warnings

TheJavaGuy:
Proof of all warnings gone:
image

Proof of warnings before this PR:
image

Note: I only fixed branch 1.5 because that was the only branch with WARN.

### DIFF
--- a/docs/modules/ROOT/pages/includes/quarkus-cxf.adoc
+++ b/docs/modules/ROOT/pages/includes/quarkus-cxf.adoc
@@ -604,22 +604,6 @@ endif::add-copy-button-to-env-var[]
 |
 
 
-a| [[quarkus-cxf_quarkus.cxf.client.-clients-.service-interface]]`link:#quarkus-cxf_quarkus.cxf.client.-clients-.service-interface[quarkus.cxf.client."clients".service-interface]`
-
-[.description]
---
-The client interface class
-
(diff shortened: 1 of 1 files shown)


ppalaga:
I am afraid, that changed file is generated. Rebuilding the whole source tree will revert your changes. One would have to fix the Quarkus tool that generates the file. I think it is not worth doing for this old branch.

Thanks for the contribution anyway!

TheJavaGuy:
@ppalaga Are you sure about that? I've done about a dozen similar PRs in other extensions and this is the first time I hear that `.adoc` files are generated. If they're generated, where's the source for them then?

Another thing, if branch 1.5 is old how come the documentation for it is still built in quarkiverse?

ppalaga:
> this is the first time I hear that .adoc files are generated.

In 1.5 only this single .adoc file is generated.

> If they're generated, where's the source for them then?

The generated .adoc file is just a table of config options supported by quarkus-cxf. Those are gathered from AST during the compilation. It is implemented inside `quarkus-extension-processor` plugged into the compiler around here https://github.com/quarkiverse/quarkus-cxf/blob/main/extensions/core/runtime/pom.xml#L118-L122

The quality of the .adoc output got much better recently and that's perhaps why you do not see those warnings in newer branches.

Issue #973, opened 2023-08-19
Netty Http2Exception: Flow control window exceeded for stream: 0 when sending a ~64 KiB attachment


ppalaga:
Originally reported as https://github.com/quarkiverse/quarkus-cxf/discussions/971 by @xperjon

There are two slightly different ways to reproduce the issue. Both are present in https://github.com/quarkiverse/quarkus-cxf/pull/974

1. Send an attachment over 64 KiB - see MtomTest.largeAttachment() in #971 
2. Soak the endpoint with requests slightly under 64 KiB - see MtomTest.soak() in #971 

In both cases, the stack trace (on the server side) is the following: 

```
WARN  [org.apa.cxf.pha.PhaseInterceptorChain] (executor-thread-1) Interceptor for {https://quarkiverse.github.io/quarkiverse-docs/quarkus-cxf/test/mtom}MtomService has thrown exception, unwinding now: java.lang.RuntimeException: java.io.IOException: java.io.IOException: io.netty.handler.codec.http2.Http2Exception: Flow control window exceeded for stream: 0
        at org.apache.cxf.attachment.LazyAttachmentCollection$1.hasNext(LazyAttachmentCollection.java:108)
        at org.apache.cxf.binding.soap.saaj.SAAJInInterceptor.handleMessage(SAAJInInterceptor.java:209)
        at org.apache.cxf.jaxws.handler.soap.SOAPMessageContextImpl.getMessage(SOAPMessageContextImpl.java:79)
        at org.apache.cxf.jaxws.handler.soap.SOAPHandlerInterceptor.getOpQName(SOAPHandlerInterceptor.java:318)
        at org.apache.cxf.jaxws.handler.AbstractJAXWSHandlerInterceptor.setupBindingOperationInfo(AbstractJAXWSHandlerInterceptor.java:110)
        at org.apache.cxf.jaxws.handler.soap.SOAPHandlerInterceptor.createProtocolMessageContext(SOAPHandlerInterceptor.java:253)
        at org.apache.cxf.jaxws.handler.soap.SOAPHandlerInterceptor.handleMessageInternal(SOAPHandlerInterceptor.java:158)
        at org.apache.cxf.jaxws.handler.soap.SOAPHandlerInterceptor.handleMessage(SOAPHandlerInterceptor.java:121)
        at org.apache.cxf.jaxws.handler.soap.SOAPHandlerInterceptor.handleMessage(SOAPHandlerInterceptor.java:70)
[...]

PR #974, opened 2023-08-19
Reproducer for #973 Http2Exception: Flow control window exceeded for stream: 0 when sending a ~64 KiB attachment

ppalaga:
Reproducer for #973


ppalaga:
Replaced by https://github.com/quarkiverse/quarkus-cxf/pull/993

PR #990, opened 2023-08-30
Allow configuration of proxy authentication via application.properties

### LINKED ISSUE #987: Revisit Advanced SOAP client topics docs page after PR #986
PR https://github.com/quarkiverse/quarkus-cxf/pull/986 introduced some new properties, so we should add/change our recommendations on https://docs.quarkiverse.io/quarkus-cxf/dev/user-guide/advanced-soap-client-topics.html

### LINKED ISSUE #988: Allow configuration of proxy authentication via application.properties
As requested by @ppalaga in https://github.com/quarkiverse/quarkus-cxf/pull/986#issuecomment-1699341406.

We can make the [proxy authentication](https://cxf.apache.org/docs/client-http-transport-including-ssl-support.html#ClientHTTPTransport(includingSSLsupport)-ProxyAuthentication) properties configurable via `application.properties`.  I haven't thought through this - but wondering if we can also optionally allow setting a system property for HTTPS endpoints, see https://github.com/quarkiverse/quarkus-cxf/discussions/941.

> This works over HTTPS and HTTPS, but note for the latter it is necessary to set the following system property (see [here](https://www.oracle.com/technetwork/java/javase/8u111-relnotes-3124969.html) for more information "Disable Basic authentication for HTTPS tunneling"):

> `-Djdk.http.auth.tunneling.disabledSchemes=`

ppalaga:
fix #988
fix #987

### DIFF
--- a/docs/modules/ROOT/examples/calculator-client/application.properties
+++ b/docs/modules/ROOT/examples/calculator-client/application.properties
@@ -45,3 +45,10 @@ quarkus.cxf.client.clientWithRuntimeInitializedPayload.native.runtime-initialize
 quarkus.native.additional-build-args = --initialize-at-run-time=io.quarkiverse.cxf.client.it.rtinit.Operands\\,io.quarkiverse.cxf.client.it.rtinit.Result
 
 quarkus.cxf.codegen.wsdl2java.includes = wsdl/*.wsdl
+
+quarkus.cxf.client.proxiedCalculator.wsdl = wsdl/CalculatorService.wsdl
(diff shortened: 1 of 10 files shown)


ppalaga:
I have no idea how to set up a proxy using testcontainers and whether it is possible at all.

shumonsharif:
Maybe something like this would work?  https://hub.docker.com/r/tarampampam/3proxy/

ppalaga:
Yeah, thanks for the tip, @shumonsharif , however, 3proxy is not very flexible when it comes to resolving the proxied host names. It accepts only proper external DNS servers, which is far from trivial to set up in Docker. I ended up with `mitmproxy/mitmproxy`. It resolves docker network aliases OOTB.

PR #996, opened 2023-09-04
Disable GH actions upgrades by Dependabot

ppalaga:
... because they are also handled by Renovate


ppalaga:
The failure in MTOM test is unrelated

PR #1019, opened 2023-09-15
Injected CXFClientInfo does not reflect the wsdl and client-endpoint-url set in application.properties

### LINKED ISSUE #491: Injected CXFClientInfo does not reflect the wsdl and client-endpoint-url set in application.properties
I am about to add a test case in an upcoming PR.

Another related issue is that it currently does not seem to be possible to inject two separate CXFClientInfos that implement the same service interface, but point at different client-endpoint-urls. I wonder whether this can be implemented at all?

ppalaga:
fix #491


extensions/core/runtime/src/main/java/io/quarkiverse/cxf/CxfClientProducer.java
```diff
@@ -94,7 +101,7 @@ private Object produceCxfClient(CXFClientInfo cxfClientInfo) {
         QuarkusClientFactoryBean quarkusClientFactoryBean = new QuarkusClientFactoryBean();
         QuarkusJaxWsProxyFactoryBean factory = new QuarkusJaxWsProxyFactoryBean(quarkusClientFactoryBean, interfaces);
         factory.setServiceClass(seiClass);
-        LOGGER.debugf("using servicename {%s}%s", cxfClientInfo.getWsNamespace(), cxfClientInfo.getWsName());
+        LOGGER.warnf("using servicename {%s}%s", cxfClientInfo.getWsNamespace(), cxfClientInfo.getWsName());
```
famod, line 104: @ppalaga just noticed this output in 2.4.0 - looks like an oversight?

famod: See #1029

ppalaga: Thanks, @famod! It is definitely a left over from my experiments.

PR #1034, opened 2023-09-20
Add Contributor guide and Release guide

ppalaga:


docs/modules/ROOT/pages/contributor-guide/index.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+[[how-to-build]]
+== How to build
+
+Checkout the code
+
+[source,shell]
+----
+$ git clone https://github.com/quarkiverse/quarkus-cxf.git -o upstream
+$ cd camel-quarkus
```
famod, line 25:
```suggestion
$ cd quarkus-cxf
```
...no?

famod:
I was wondering whether we should add something to prevent errors like the following after every release:
https://github.com/quarkiverse/quarkus-cxf/actions/runs/6220354779/job/16880212982
First I was about to add something to the `release` gh action but then again: why is `[calculator-client](https://github.com/quarkiverse/quarkus-cxf/tree/main/docs/modules/ROOT/examples/calculator-client)` checked in anyway?
I mean there is a copy goal in the actual IT module so it's more of a build result than actual source, no?

docs/modules/ROOT/pages/contributor-guide/releasing.adoc
```diff
@@ -0,0 +1,54 @@
...
+$ cd generated-platform-project
+$ ../mvnw install -pl quarkus/bom,quarkus-cxf/bom,quarkus-universe/bom -am
+# Make sure the generated tests pass
+$ cd quarkus-cxf/integration-tests
+$ ../../../mvnw verify -Pnative
+# If all is green, commit changes and send a PR
+$ git checkout -b upgrade-qcxf-$NEW_VERSION
+$ git add -A
+$ git commit -m "Upgrade to Quarkus CXF $NEW_VERSION"
+$ git push your-fork upgrade-qcxf-$NEW_VERSION
```
famod, line 50: Well, couldn't we add a script for all that too?

docs/modules/ROOT/pages/contributor-guide/index.adoc
ppalaga: Thanks, of course, a copy-and-paste error

docs/modules/ROOT/pages/contributor-guide/releasing.adoc
ppalaga: I have not seen such a need so far. Perhaps because some of the steps require manual verification. But yes a big part could be scripted and maybe also added to the release GHA workflow. The question is whether awaiting the released artifacts to appear on Maven Central would not took too long. It is scriptable for sure.

ppalaga:
> why is [calculator-client](https://github.com/quarkiverse/quarkus-cxf/tree/main/docs/modules/ROOT/examples/calculator-client) checked in anyway?
> I mean there is a copy goal in the actual IT module so it's more of a build result than actual source, no?

There are snippets cited from those copied files in the docs. At the same time Antora is very opinionated where the snippet sources have to be located. There is no way to refence a file outside of the docs/modules folder.

famod:
> There is no way to refence a file outside of the docs/modules folder.

Ok, but what I have in mind is to gitignore `docs/modules/ROOT/examples/calculator-client`.
Then either the docs module is copying stuff over from the IT module or it stays like now (but then the docs module is less self-contained).

PS: This is all under assumption that the actual docs generation is done via `AntoraTest` which takes part in the regular Maven lifecycle.

famod:
Ah sure, totally forgot about the quarkiverse bits.

PR #1049, opened 2023-09-27
Allow configuring LoggingFeature through application.properties

### LINKED ISSUE #595: Allow configuring `LoggingFeature` through `application.properties`
The [docs](https://quarkiverse.github.io/quarkiverse-docs/quarkus-cxf/dev/message-logging.html) say that one has to write a producer method to get a `LoggingFeature` with the desired config.

While this works just fine, it seems that the logging module could provide a producer OOTB (for more developer joy), configured via a set of properties in the quarkus-cxf namespace.
The icing on the cake would be if you could create multiple features (with different settings) purely via config, e.g.:
```
quarkus.cxf.logging."logging-feature-1".sensitive-element-names=foo,bar
quarkus.cxf.logging."logging-feature-2".sensitive-element-names=baz,bing
```
You'd then activate the desired feature via e.g.:
```
quarkus.cxf.client."client-1".features=logging-feature-1
quarkus.cxf.client."client-2".features=logging-feature-2
```
The above would require an extension to how we look up features (new: via `@Named`).

ppalaga:
fix #595


ppalaga:
@famod 

here is the idea:
image

I hope it fulfills your expectations?

famod:
Very nice, thanks!

The icing on the cake would be to configure a logging feature once and reuse it for muliple clients, but I guess that'll be a future improvement.

ppalaga:
f205870:

* Added a better workaround for https://github.com/quarkusio/quarkus/issues/36189

PR #1092, opened 2023-11-03
Document X-Forwarded support 

### LINKED ISSUE #556: Document X-Forwarded support
We support rewriting the location value in WSDL through X-Forwarded headers -  see the [XForwardedHeadersTest](https://github.com/quarkiverse/quarkus-cxf/blob/main/integration-tests/server/src/test/java/io/quarkiverse/cxf/it/server/XForwardedHeadersTest.java)
This feature needs to be documented, including at least the following topics:
* A list of supported X-Forwarded headers and what they do
* A rough overview, how the rewriting is implemented
* The role of `quarkus.http.proxy.enable-forwarded-host`, `quarkus.http.proxy.enable-forwarded-prefix` and             `quarkus.http.proxy.proxy-address-forwarding`configuration properties

~~CXF on WildFly supports some advanced address rewriting for the usage with load balancers or proxies: https://docs.wildfly.org/24/Developer_Guide.html#address-rewrite I wonder whether we need to have something like that?~~

ppalaga:
fix #556

### DIFF
--- a/docs/modules/ROOT/pages/user-guide/advanced-soap-server-topics.adoc
+++ b/docs/modules/ROOT/pages/user-guide/advanced-soap-server-topics.adoc
@@ -1,0 +1,63 @@
+= Advanced SOAP server topics
+
+== Running behind a reverse proxy
+
+SOAP requests aimed towards services running on Quarkus can be routed through proxies that generate additional headers
+(e.g. `X-Forwarded-Host`) to keep information from the client-facing side of the proxy servers that is altered or lost when they are involved.
+In those scenarios, Quarkus can be configured to automatically update information like protocol, host, port and URI reflecting the values in these headers.
+
+TIP: Refer to https://quarkus.io/guides/http-reference#reverse-proxy[Quarkus HTTP reference] for more details.
+
+{quarkus-cxf-project-name} support for various `X-Forwarded` headers works in line with Quarkus configuration.
+
+IMPORTANT: Activating this feature leaves the server exposed to several security issues (i.e. information spoofing). Consider activating it only when running behind a reverse proxy.
+
+These are the relevant Quarkus properties and their effect on {quarkus-cxf-project-name}:
+
+* `https://quarkus.io/guides/all-config#quarkus-vertx-http_quarkus.http.proxy.proxy-address-forwarding[quarkus.http.proxy.proxy-address-forwarding]` - the main switch to enable the rewriting of the request destination parts.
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/user-guide/advanced-soap-server-topics.adoc
```diff
@@ -0,0 +1,63 @@
+= Advanced SOAP server topics
+
+== Running behind a reverse proxy
+
+SOAP requests aimed towards services running on Quarkus can be routed through proxies that generate additional headers
+(e.g. `X-Forwarded-Host`) to keep information from the client-facing side of the proxy servers that is altered or lost when they are involved.
+In those scenarios, Quarkus can be configured to automatically update information like protocol, host, port and URI reflecting the values in these headers.
+
+TIP: Refer to https://quarkus.io/guides/http-reference[Quarkus HTTP reference] for more details.
```
shumonsharif, line 9: Did you intend to reference:  https://quarkus.io/guides/http-reference#reverse-proxy ?

docs/modules/ROOT/pages/user-guide/advanced-soap-server-topics.adoc
```diff
@@ -0,0 +1,63 @@
...
+
+{quarkus-cxf-project-name} support for various `X-Forwarded` headers works in line with Quarkus configuration.
+
+IMPORTANT: Activating this feature leaves the server exposed to several security issues (i.e. information spoofing). Consider activate it only when running behind a reverse proxy.
```
shumonsharif, line 13: Consider **_activating_** it only when .... (looks like an issue with the Quarkus docs as well).

shumonsharif (commented):
Hi @ppalaga This looks great; just the couple minor comments ... thanks for putting it together!

docs/modules/ROOT/pages/user-guide/advanced-soap-server-topics.adoc
ppalaga: No I didn't but the URL with the anchor is much better :-)

docs/modules/ROOT/pages/user-guide/advanced-soap-server-topics.adoc
ppalaga: Let's make our copy better than their original :D

ppalaga:
4758dfa: addressed the feedback of @shumonsharif

PR #1102, opened 2023-11-08
Global logging feature, deprecate the logging extension and move its functionality to the core extension

### LINKED ISSUE #270: Global features
AFAICS, you cannot register a "global" feature for all clients via e.g.:
```
quarkus.cxf.client.features=org.apache.cxf.feature.LoggingFeature
```
(or `quarkus.cxf.endpoint.features` for that matter)

For a DRY and KISS config approach, registering global features should be supported (and probably interceptors etc.).

On question to answer would be whether client/endpoint specific settings are _adding_ to the global ones or are _replacing_ them instead (for the respective client/endpoint).

ppalaga:
Fix #270


ppalaga:
This is quite a major change. @shumonsharif @famod, would you like to express your opinion, whether this is a good idea?

ppalaga:
Any chance to review or (n)ack?

ppalaga:
Thanks @shumonsharif , those are good ideas, I have filed https://github.com/quarkiverse/quarkus-cxf/issues/1107

famod:
@ppalaga first off, I finally answered you initial question: https://github.com/quarkiverse/quarkus-cxf/issues/270#issuecomment-1810063284
And of course thanks for addressing this topic!

One question: That `quarkus.cxf.logging.enabled-for = clients|services|both|none` idea didn't make it in the end, right?
So you can only enable/disable logging for clients and services globally, but you can disable/enable for clients or services via two distinct properties or even per client or service, right?

ppalaga:
> One question: That `quarkus.cxf.logging.enabled-for = clients|services|both|none` idea didn't make it in the end, right?

No, because I forgot 😂. I can implement it if you find it useful?

> So you can only enable/disable logging for clients and services globally, but you can disable/enable for clients or services via two distinct properties or even per client or service, right?

Yes, with current impl, if you enable logging globally, you can still mute it for individual clients and services by passing `quarkus.cxf.[client.myClient|endpoint."/foo"].logging.enabled = false`

ppalaga:
3c0e443: Added `quarkus.cxf.logging.enabled-for = clients|services|clients-and-services|none`.

Issue #1066, opened 2023-10-06
Large responses written slowly due to allocating too small buffers


fmarchioni:
I have found evidence of slow performance of Quarkus CXF extension.
In a nutshell, a single invocation of a SOAP endpoint takes about 10 seconds to return a response. The performance further degrades when concurrent calls are made.

From my initial analysis, I can see that the main Thread that replies to your request interacts with Saxon parsing library for most of the time, until the stream of data is returned to the customer:
```
"vert.x-worker-thread-0" #114 prio=5 os_prio=0 cpu=8272.53ms elapsed=222.58s tid=0x00007f438c0daa70 nid=0x2d1a runnable  [0x00007f44fb2fa000]
   java.lang.Thread.State: RUNNABLE
at com.sun.org.apache.xerces.internal.dom.ParentNode.nodeListItem(java.xml@17.0.2/ParentNode.java:766)
at com.sun.org.apache.xerces.internal.dom.ParentNode.item(java.xml@17.0.2/ParentNode.java:797)
at net.sf.saxon.dom.DOMNodeWrapper$ChildEnumeration.skipFollowingTextNodes(DOMNodeWrapper.java:1122)
at net.sf.saxon.dom.DOMNodeWrapper$ChildEnumeration.next(DOMNodeWrapper.java:1159)
- locked <0x000000045182aaa8> (a com.sun.org.apache.xerces.internal.dom.DeferredDocumentImpl)
at net.sf.saxon.tree.util.Navigator$EmptyTextFilter.next(Navigator.java:1292)
at net.sf.saxon.tree.util.Navigator$EmptyTextFilter.next(Navigator.java:1274)
at net.sf.saxon.tree.iter.LookaheadIteratorImpl.next(LookaheadIteratorImpl.java:53)
at net.sf.saxon.tree.jiter.WrappingJavaIterator.next(WrappingJavaIterator.java:48)
at net.sf.saxon.tree.jiter.WrappingJavaIterator.next(WrappingJavaIterator.java:24)
at net.sf.saxon.tree.util.Navigator.copy(Navigator.java:695)
. . . .
"vert.x-worker-thread-0" #114 prio=5 os_prio=0 cpu=15308.80ms elapsed=230.93s tid=0x00007f438c0daa70 nid=0x2d1a runnable  [0x00007f44fb2fc000]
   java.lang.Thread.State: RUNNABLE
[...]

PR #1105, opened 2023-11-10
Use directBuffer with 8m initial capacity 

ppalaga:
#1066

### DIFF
--- a/extensions/core/runtime/src/main/java/io/quarkiverse/cxf/transport/VertxServletOutputStream.java
+++ b/extensions/core/runtime/src/main/java/io/quarkiverse/cxf/transport/VertxServletOutputStream.java
@@ -98,7 +98,7 @@ public class VertxServletOutputStream extends ServletOutputStream {
         ByteBuf buffer = pooledBuffer;
         try {
             if (buffer == null) {
-                pooledBuffer = buffer = PooledByteBufAllocator.DEFAULT.directBuffer();
+                pooledBuffer = buffer = PooledByteBufAllocator.DEFAULT.directBuffer(8 * 1024 * 1024);
             }
             while (rem > 0) {
                 int toWrite = Math.min(rem, buffer.writableBytes());
@@ -107,7 +107,7 @@ public class VertxServletOutputStream extends ServletOutputStream {
                 idx += toWrite;
                 if (!buffer.isWritable()) {
                     ByteBuf tmpBuf = buffer;
-                    this.pooledBuffer = buffer = PooledByteBufAllocator.DEFAULT.directBuffer();
+                    this.pooledBuffer = buffer = PooledByteBufAllocator.DEFAULT.directBuffer(8 * 1024 * 1024);
                     writeBlocking(tmpBuf, false);
                 }
             }
@@ -147,7 +147,7 @@ public class VertxServletOutputStream extends ServletOutputStream {
         //do all this in the same lock
         synchronized (LOCK) {
             try {
-                boolean bufferRequired = awaitWriteable() || (overflow != null && overflow.size() > 0);
+                boolean bufferRequired = (overflow != null && overflow.size() > 0) || awaitWriteable();
                 if (bufferRequired) {
                     //just buffer everything
                     registerDrainHandler();


franz1981 (changes requested):
This can be equally dangerous (especially if the write happen in the blocking thread pool executor), see https://github.com/quarkusio/quarkus/issues/32546.

I strongly suggest to try reuse https://github.com/quarkusio/quarkus/pull/32858 (AppendBuffer and maybe the stream itself?) and expose both min and buffer size, to both save number of packets creations AND to avoid duplicated code

wdyt?

geoand:
Definitely +1 to using the same approach as @franz1981 introduced in RESTEasy Reactive as if has worked out really well for us

franz1981:
Just fyi (to be verified): https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java#L122 32K is the max cacheable size for pooled allocations, which means that 8 mb will not be pooled AND could create other problems eg https://sourceware.org/bugzilla/show_bug.cgi?id=11261

Tldr performing malloc/free (used under the hood by Netty for unspoiled big allocations) can cause RSS increases due to thread-local arenas of glibc malloc.

franz1981:
@ppalaga can you share here the stack trace which lead to VertxServletOutputStream's method call?
Maybe there's some interesting low hanging fruit which would avoid other intermediate big allocation(s) as well, thanks to Netty's ability to directly encode char sequence(s) into the pooled buffers
(and/or leverage to some optimization I've done elsewhere, like -> https://github.com/undertow-io/undertow/pull/1424).

ppalaga:
> can you share here the stack trace which lead to VertxServletOutputStream's method call?

Yep, here it is:

```
java.lang.RuntimeException
     at io.quarkiverse.cxf.transport.VertxServletOutputStream.awaitWriteable(VertxServletOutputStream.java:185)
     at io.quarkiverse.cxf.transport.VertxServletOutputStream.write(VertxServletOutputStream.java:150)
     at io.quarkiverse.cxf.transport.VertxServletOutputStream.writeBlocking(VertxServletOutputStream.java:124)
     at io.quarkiverse.cxf.transport.VertxServletOutputStream.write(VertxServletOutputStream.java:111)
     at org.apache.cxf.io.AbstractWrappedOutputStream.write(AbstractWrappedOutputStream.java:51)
     at com.ctc.wstx.io.UTF8Writer.write(UTF8Writer.java:143)
     at com.ctc.wstx.sw.BufferingXmlWriter.flushBuffer(BufferingXmlWriter.java:1417)
     at com.ctc.wstx.sw.BufferingXmlWriter.writeRaw(BufferingXmlWriter.java:256)
     at com.ctc.wstx.sw.BufferingXmlWriter.writeCharacters(BufferingXmlWriter.java:600)
     at com.ctc.wstx.sw.BaseStreamWriter.writeCharacters(BaseStreamWriter.java:467)
     at org.apache.cxf.staxutils.StaxUtils.copy(StaxUtils.java:759)
     at org.apache.cxf.staxutils.StaxUtils.copy(StaxUtils.java:707)
     at org.apache.cxf.jaxws.handler.logical.LogicalHandlerOutInterceptor$LogicalHandlerOutEndingInterceptor.handleMessage(LogicalHandlerOutInterceptor.java:183)
     at org.apache.cxf.phase.PhaseInterceptorChain.doIntercept(PhaseInterceptorChain.java:307)
     at org.apache.cxf.interceptor.OutgoingChainInterceptor.handleMessage(OutgoingChainInterceptor.java:90)
     at org.apache.cxf.phase.PhaseInterceptorChain.doIntercept(PhaseInterceptorChain.java:307)
     at org.apache.cxf.transport.ChainInitiationObserver.onMessage(ChainInitiationObserver.java:121)
     at org.apache.cxf.transport.http.AbstractHTTPDestination.invoke(AbstractHTTPDestination.java:265)
     at org.apache.cxf.transport.servlet.ServletController.invokeDestination(ServletController.java:233)
     at org.apache.cxf.transport.servlet.ServletController.invoke(ServletController.java:207)
     at org.apache.cxf.transport.servlet.ServletController.invoke(ServletController.java:159)
     at io.quarkiverse.cxf.transport.CxfHandler.process(CxfHandler.java:219)
     at io.quarkiverse.cxf.transport.CxfHandler.handle(CxfHandler.java:156)
     at io.quarkiverse.cxf.transport.CxfHandler.handle(CxfHandler.java:40)
     at io.vertx.ext.web.impl.BlockingHandlerDecorator.lambda$handle$0(BlockingHandlerDecorator.java:48)
     at io.vertx.core.impl.ContextBase.lambda$executeBlocking$1(ContextBase.java:180)
     at io.vertx.core.impl.ContextInternal.dispatch(ContextInternal.java:277)
     at io.vertx.core.impl.ContextBase.lambda$internalExecuteBlocking$2(ContextBase.java:199)
     at io.quarkus.vertx.core.runtime.VertxCoreRecorder$14.runWith(VertxCoreRecorder.java:582)
```

franz1981:
Thanks @ppalaga so...quick analysis:
https://github.com/FasterXML/woodstox/blob/8b49745c1277ef75d880782ee92cf336a9c15395/src/main/java/com/ctc/wstx/io/UTF8Writer.java#L143  "seems" to suggest that we cannot "steal" the `byte[]`  in https://github.com/FasterXML/woodstox/blob/8b49745c1277ef75d880782ee92cf336a9c15395/src/main/java/com/ctc/wstx/io/UTF8Writer.java#L50 given that's reused to perform UTF-8 encoding (it's the same for Json encoding in Jackson actually - sameol' sameol').

Due to this, I strongly recommend to try reuse the existing jackson approach mentioned earlier, because it's both fixing the issue and preventing further performance issues (till the next one :P).

ppalaga:
> Just fyi (to be verified): https://github.com/netty/netty/blob/4.1/buffer/src/main/java/io/netty/buffer/PooledByteBufAllocator.java#L122 32K is the max cacheable size for pooled allocations, which means that 8 mb will not be pooled AND could create other problems eg https://sourceware.org/bugzilla/show_bug.cgi?id=11261

Thanks for the hint, but the idea with 8m was not meant for merging. It was rather a quick fix to figure out whether the approach brings any improvement at all. 

> I strongly recommend to try reuse the existing jackson approach mentioned earlier

You mean the AppendBuffer? I created a new PR https://github.com/quarkiverse/quarkus-cxf/pull/1106 Let's continue the discussion there.

Issue #1173, opened 2024-01-11
Make algorithm suites configurable through application.properties thus allowing FIPS compliant suites


JiriOndrusek:
To make quarkus-cxf works correctly o FIPS, custom algorithm suite has to be configured.

It would be nice to have an option to configure such suite via properties, where default values are working on FIPS.

Modules integration-tests/ws-trust and integration-tests/ws-security-policy should work on FIPS with this change.

ppalaga:
For the record, we agreed to try to propose the algorithm configuration on CXF level https://issues.apache.org/jira/browse/CXF-8971

ppalaga:
Fixed by 96405020012ec2c95e537dd7392df4e067e1ca95

PR #1174, opened 2024-01-11
Customizable algorithm suite defaulting to FIPS compliance

JiriOndrusek:
fixes https://github.com/quarkiverse/quarkus-cxf/issues/1173


JiriOndrusek:
The ws-trust native failure is a a bug in my code and I'll take care of it.

JiriOndrusek:
@ppalaga @ffang 
I'd like to ask for help with the native failure.
I think that I see, why the test is failing, but I'm not sure how to fix it correctly.

Let me explain (native mode):

1. During the build, the map [AlgorithmSuite.ALGORITHM_SUITE_TYPES](https://github.com/apache/ws-wss4j/blob/wss4j-3.0.2/policy/src/main/java/org/apache/wss4j/policy/model/AlgorithmSuite.java#L34) - it is created because it is static field (java identifier of the map is **1382497575**)
2. The custom alghoritm suite is created in the recorder (part of this PR) in the static init into **the same** map  (java identifier of the map is **1382497575**)
3. During test execution **the new** map is created
```
2024-01-12 10:33:45,915 INFO  [org.apa.cxf.wsd.ser.fac.ReflectionServiceFactoryBean] (main) Creating Service {https://quarkiverse.github.io/quarkiverse-docs/quarkus-cxf/test/ws-trust}TrustHelloService from WSDL: http://localhost:8081/services/jaxws-samples-wsse-policy-trust/TrustHelloService?wsdl
(java identifier of the map is **888350069**)
```

  4. The new map does not contain the custom alg. suite, therefore the execution fails.

I saved the stacktrace of the second map cration:
```
        at org.apache.wss4j.policy.model.AlgorithmSuite.<clinit>(AlgorithmSuite.java:34)
        at org.apache.cxf.ws.security.policy.WSSecurityPolicyLoader.registerBuilders(WSSecurityPolicyLoader.java:275)
        at org.apache.cxf.ws.security.policy.WSSecurityPolicyLoader.<init>(WSSecurityPolicyLoader.java:106)
        at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
        at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:480)
        at org.apache.cxf.bus.extension.Extension.load(Extension.java:218)
        at org.apache.cxf.bus.extension.ExtensionManagerImpl.loadAndRegister(ExtensionManagerImpl.java:198)
        at org.apache.cxf.bus.extension.ExtensionManagerImpl.getBeansOfType(ExtensionManagerImpl.java:333)
        at org.apache.cxf.ws.policy.AssertionBuilderRegistryImpl.loadDynamic(AssertionBuilderRegistryImpl.java:95)
        at org.apache.neethi.AssertionBuilderFactoryImpl.build(AssertionBuilderFactoryImpl.java:109)
        at org.apache.neethi.PolicyBuilder.processOperationElement(PolicyBuilder.java:224)
        at org.apache.neethi.PolicyBuilder.getAllOperator(PolicyBuilder.java:184)
        at org.apache.neethi.PolicyBuilder.processOperationElement(PolicyBuilder.java:217)
        at org.apache.neethi.PolicyBuilder.getExactlyOneOperator(PolicyBuilder.java:180)
        at org.apache.neethi.PolicyBuilder.processOperationElement(PolicyBuilder.java:215)
        at org.apache.neethi.PolicyBuilder.getPolicyOperator(PolicyBuilder.java:174)
        at org.apache.neethi.PolicyBuilder.getPolicy(PolicyBuilder.java:124)
        at org.apache.cxf.ws.policy.attachment.wsdl11.Wsdl11AttachmentPolicyProvider.getElementPolicy(Wsdl11AttachmentPolicyProvider.java:197)
        at org.apache.cxf.ws.policy.attachment.wsdl11.Wsdl11AttachmentPolicyProvider.getElementPolicy(Wsdl11AttachmentPolicyProvider.java:170)
        at org.apache.cxf.ws.policy.attachment.wsdl11.Wsdl11AttachmentPolicyProvider.getElementPolicy(Wsdl11AttachmentPolicyProvider.java:163)
        at org.apache.cxf.ws.policy.attachment.wsdl11.Wsdl11AttachmentPolicyProvider.getEffectivePolicy(Wsdl11AttachmentPolicyProvider.java:77)
        at org.apache.cxf.ws.policy.PolicyEngineImpl.getAggregatedServicePolicy(PolicyEngineImpl.java:461)
        at org.apache.cxf.ws.policy.EndpointPolicyImpl.initializePolicy(EndpointPolicyImpl.java:151)
        at org.apache.cxf.ws.policy.EndpointPolicyImpl.initialize(EndpointPolicyImpl.java:140)
        at org.apache.cxf.ws.policy.PolicyEngineImpl.createEndpointPolicyInfo(PolicyEngineImpl.java:614)
        at org.apache.cxf.ws.policy.PolicyEngineImpl.getEndpointPolicy(PolicyEngineImpl.java:326)
        at org.apache.cxf.ws.policy.PolicyEngineImpl.getClientEndpointPolicy(PolicyEngineImpl.java:313)
        at org.apache.cxf.ws.policy.PolicyDataEngineImpl.getClientEndpointPolicy(PolicyDataEngineImpl.java:61)
        at org.apache.cxf.transport.http.HTTPConduit.updateClientPolicy(HTTPConduit.java:372)
        at org.apache.cxf.transport.http.HTTPConduit.updateClientPolicy(HTTPConduit.java:392)
        at org.apache.cxf.transport.http.HTTPConduit.getClient(HTTPConduit.java:938)
        at org.apache.cxf.transport.http.HTTPConduit.configureConduitFromEndpointInfo(HTTPConduit.java:414)
        at org.apache.cxf.transport.http.HTTPConduit.finalizeConfig(HTTPConduit.java:494)
        at org.apache.cxf.transport.http.HTTPTransportFactory.getConduit(HTTPTransportFactory.java:257)
        at org.apache.cxf.binding.soap.SoapTransportFactory.getConduit(SoapTransportFactory.java:226)
        at org.apache.cxf.binding.soap.SoapTransportFactory.getConduit(SoapTransportFactory.java:233)
        at org.apache.cxf.endpoint.AbstractConduitSelector.createConduit(AbstractConduitSelector.java:144)
        at org.apache.cxf.endpoint.AbstractConduitSelector.getSelectedConduit(AbstractConduitSelector.java:108)
        at org.apache.cxf.endpoint.UpfrontConduitSelector.prepare(UpfrontConduitSelector.java:63)
        at org.apache.cxf.endpoint.ClientImpl.prepareConduitSelector(ClientImpl.java:885)
        at org.apache.cxf.endpoint.ClientImpl.doInvoke(ClientImpl.java:523)
        at org.apache.cxf.endpoint.ClientImpl.invoke(ClientImpl.java:439)
        at org.apache.cxf.endpoint.ClientImpl.invoke(ClientImpl.java:354)
        at org.apache.cxf.endpoint.ClientImpl.invoke(ClientImpl.java:312)
        at org.apache.cxf.frontend.ClientProxy.invokeSync(ClientProxy.java:96)
        at org.apache.cxf.jaxws.JaxWsClientProxy.invoke(JaxWsClientProxy.java:140)
        at jdk.proxy2/jdk.proxy2.$Proxy91.sayHello(Unknown Source)
        at io.quarkiverse.cxf.it.ws.trust.CxfWsTrustTest.programmaticSts(CxfWsTrustTest.java:103)

```

In the jvm execution, there is only 1 map (created by the recorder) and all works.

I'm a little bit confused, why is the recorder adding values to the statically created map but the client from the test creates anew default map? (and  thus ignoring the custom modification)

JiriOndrusek:
Per

> @ppalaga @ffang I'd like to ask for help with the native failure. I think that I see, why the test is failing, but I'm not sure how to fix it correctly.
> 
> Let me explain (native mode):
> 
> 1. During the build, the map [AlgorithmSuite.ALGORITHM_SUITE_TYPES](https://github.com/apache/ws-wss4j/blob/wss4j-3.0.2/policy/src/main/java/org/apache/wss4j/policy/model/AlgorithmSuite.java#L34) - it is created because it is static field (java identifier of the map is **1382497575**)
> 2. The custom alghoritm suite is created in the recorder (part of this PR) in the static init into **the same** map  (java identifier of the map is **1382497575**)
> 3. During test execution **the new** map is created
> 
> ```
> 2024-01-12 10:33:45,915 INFO  [org.apa.cxf.wsd.ser.fac.ReflectionServiceFactoryBean] (main) Creating Service {https://quarkiverse.github.io/quarkiverse-docs/quarkus-cxf/test/ws-trust}TrustHelloService from WSDL: http://localhost:8081/services/jaxws-samples-wsse-policy-trust/TrustHelloService?wsdl
> (java identifier of the map is **888350069**)
> ```
> 
> 4. The new map does not contain the custom alg. suite, therefore the execution fails.
> 
> I saved the stacktrace of the second map cration:
> 
> ```
>         at org.apache.wss4j.policy.model.AlgorithmSuite.<clinit>(AlgorithmSuite.java:34)
>         at org.apache.cxf.ws.security.policy.WSSecurityPolicyLoader.registerBuilders(WSSecurityPolicyLoader.java:275)
>         at org.apache.cxf.ws.security.policy.WSSecurityPolicyLoader.<init>(WSSecurityPolicyLoader.java:106)
>         at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
>         at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
>         at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
>         at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
>         at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:480)
>         at org.apache.cxf.bus.extension.Extension.load(Extension.java:218)
>         at org.apache.cxf.bus.extension.ExtensionManagerImpl.loadAndRegister(ExtensionManagerImpl.java:198)
>         at org.apache.cxf.bus.extension.ExtensionManagerImpl.getBeansOfType(ExtensionManagerImpl.java:333)
>         at org.apache.cxf.ws.policy.AssertionBuilderRegistryImpl.loadDynamic(AssertionBuilderRegistryImpl.java:95)
>         at org.apache.neethi.AssertionBuilderFactoryImpl.build(AssertionBuilderFactoryImpl.java:109)
>         at org.apache.neethi.PolicyBuilder.processOperationElement(PolicyBuilder.java:224)
>         at org.apache.neethi.PolicyBuilder.getAllOperator(PolicyBuilder.java:184)
>         at org.apache.neethi.PolicyBuilder.processOperationElement(PolicyBuilder.java:217)
>         at org.apache.neethi.PolicyBuilder.getExactlyOneOperator(PolicyBuilder.java:180)
>         at org.apache.neethi.PolicyBuilder.processOperationElement(PolicyBuilder.java:215)
>         at org.apache.neethi.PolicyBuilder.getPolicyOperator(PolicyBuilder.java:174)
>         at org.apache.neethi.PolicyBuilder.getPolicy(PolicyBuilder.java:124)
>         at org.apache.cxf.ws.policy.attachment.wsdl11.Wsdl11AttachmentPolicyProvider.getElementPolicy(Wsdl11AttachmentPolicyProvider.java:197)
>         at org.apache.cxf.ws.policy.attachment.wsdl11.Wsdl11AttachmentPolicyProvider.getElementPolicy(Wsdl11AttachmentPolicyProvider.java:170)
>         at org.apache.cxf.ws.policy.attachment.wsdl11.Wsdl11AttachmentPolicyProvider.getElementPolicy(Wsdl11AttachmentPolicyProvider.java:163)
>         at org.apache.cxf.ws.policy.attachment.wsdl11.Wsdl11AttachmentPolicyProvider.getEffectivePolicy(Wsdl11AttachmentPolicyProvider.java:77)
>         at org.apache.cxf.ws.policy.PolicyEngineImpl.getAggregatedServicePolicy(PolicyEngineImpl.java:461)
>         at org.apache.cxf.ws.policy.EndpointPolicyImpl.initializePolicy(EndpointPolicyImpl.java:151)
>         at org.apache.cxf.ws.policy.EndpointPolicyImpl.initialize(EndpointPolicyImpl.java:140)
>         at org.apache.cxf.ws.policy.PolicyEngineImpl.createEndpointPolicyInfo(PolicyEngineImpl.java:614)
>         at org.apache.cxf.ws.policy.PolicyEngineImpl.getEndpointPolicy(PolicyEngineImpl.java:326)
>         at org.apache.cxf.ws.policy.PolicyEngineImpl.getClientEndpointPolicy(PolicyEngineImpl.java:313)
>         at org.apache.cxf.ws.policy.PolicyDataEngineImpl.getClientEndpointPolicy(PolicyDataEngineImpl.java:61)
>         at org.apache.cxf.transport.http.HTTPConduit.updateClientPolicy(HTTPConduit.java:372)
>         at org.apache.cxf.transport.http.HTTPConduit.updateClientPolicy(HTTPConduit.java:392)
>         at org.apache.cxf.transport.http.HTTPConduit.getClient(HTTPConduit.java:938)
>         at org.apache.cxf.transport.http.HTTPConduit.configureConduitFromEndpointInfo(HTTPConduit.java:414)
>         at org.apache.cxf.transport.http.HTTPConduit.finalizeConfig(HTTPConduit.java:494)
>         at org.apache.cxf.transport.http.HTTPTransportFactory.getConduit(HTTPTransportFactory.java:257)
>         at org.apache.cxf.binding.soap.SoapTransportFactory.getConduit(SoapTransportFactory.java:226)
>         at org.apache.cxf.binding.soap.SoapTransportFactory.getConduit(SoapTransportFactory.java:233)
>         at org.apache.cxf.endpoint.AbstractConduitSelector.createConduit(AbstractConduitSelector.java:144)
>         at org.apache.cxf.endpoint.AbstractConduitSelector.getSelectedConduit(AbstractConduitSelector.java:108)
>         at org.apache.cxf.endpoint.UpfrontConduitSelector.prepare(UpfrontConduitSelector.java:63)
>         at org.apache.cxf.endpoint.ClientImpl.prepareConduitSelector(ClientImpl.java:885)
>         at org.apache.cxf.endpoint.ClientImpl.doInvoke(ClientImpl.java:523)
>         at org.apache.cxf.endpoint.ClientImpl.invoke(ClientImpl.java:439)
>         at org.apache.cxf.endpoint.ClientImpl.invoke(ClientImpl.java:354)
>         at org.apache.cxf.endpoint.ClientImpl.invoke(ClientImpl.java:312)
>         at org.apache.cxf.frontend.ClientProxy.invokeSync(ClientProxy.java:96)
>         at org.apache.cxf.jaxws.JaxWsClientProxy.invoke(JaxWsClientProxy.java:140)
>         at jdk.proxy2/jdk.proxy2.$Proxy91.sayHello(Unknown Source)
>         at io.quarkiverse.cxf.it.ws.trust.CxfWsTrustTest.programmaticSts(CxfWsTrustTest.java:103)
> ```
> 
> In the jvm execution, there is only 1 map (created by the recorder) and all works.
> 
> I'm a little bit confused, why is the recorder adding values to the statically created map but the client from the test creates anew default map? (and thus ignoring the custom modification)

Perhaps different bus? which has to be customized as well? (but then the execution should fail in JVM as well)

JiriOndrusek:
I forgot to put a reason of the failure here (the custom alg suite is not registered):

```
[ERROR] io.quarkiverse.cxf.it.ws.trust.CxfWsTrustIT.programmaticSts -- Time elapsed: 1.006 s <<< ERROR!
jakarta.xml.ws.soap.SOAPFaultException: Algorithm suite "CustomizedAlgorithmSuite" is not registered
        at org.apache.cxf.jaxws.JaxWsClientProxy.mapException(JaxWsClientProxy.java:195)

```

integration-tests/ws-trust/src/test/java/io/quarkiverse/cxf/it/ws/trust/CxfWsTrustTest.java
```diff
@@ -76,28 +80,28 @@ void wsdl() {
...
-            setupWsseAndSTSClient(proxy, bus, stsURL, stsServiceName, stsPortName);
-
-            Assertions.assertThat(proxy.sayHello()).isEqualTo("WS-Trust Hello World!");
-        } finally {
-            bus.shutdown(true);
-        }
+
+        BusFactory.setThreadDefaultBus(bus);
+
+        //programmatic registration of customizedAlgorithmSuite
```
JiriOndrusek, line 86: @ffang I had to register custom algorithm in this test. I suppose that in case of programmatic STS programatic registration makes sense, what do you think?

extensions/ws-security/runtime/src/main/java/io/quarkiverse/cxf/ws/security/CustomAlgSuiteLoader.java
```diff
@@ -0,0 +1,111 @@
...
+import org.apache.wss4j.policy.model.AbstractSecurityAssertion;
+import org.apache.wss4j.policy.model.AlgorithmSuite;
+import org.eclipse.microprofile.config.Config;
+
+/**
+ * This class retrieves a custom AlgorithmSuite for use with restricted security policies
+ */
+public class CustomAlgSuiteLoader implements AlgorithmSuiteLoader {
+
+    public static final String DIGEST_ALGORITHM = "http://www.w3.org/2001/04/xmlenc#sha256";
```
JiriOndrusek, line 34: @ppalaga I'd like to access default values of CxfBuilTimeConfig class here to avoid duplicated constant creation. Do you see any other way of avoiding duplicates? (I don't see any buildTimeConfig for ws-security, where it would be probably avoidlable. (core-deployment does not see ws-secuurity/runtime and vice versa)

integration-tests/ws-security-server/src/test/java/io/quarkiverse/cxf/it/ws/securitypolicy/server/CxfWssSecurityPolicyServerTest.java
```diff
@@ -191,9 +196,15 @@ void wsdl() {
...
     WssSecurityPolicyHelloService getPlainClient() {
-        return QuarkusCxfClientTestUtil.getClient(
+        WssSecurityPolicyHelloService client = QuarkusCxfClientTestUtil.getClient(
                 "https://quarkiverse.github.io/quarkiverse-docs/quarkus-cxf/ws-securitypolicy",
                 WssSecurityPolicyHelloService.class,
                 "/soap/security-policy-hello");
+
+        //programmatic registration of customizedAlgorithmSuite
+        final Config config = ConfigProvider.getConfig();
+        new CustomAlgSuiteLoader(((JaxWsClientProxy) Proxy.getInvocationHandler(client)).getClient().getBus(), config);
```
JiriOndrusek, line 206: @ffang The same issue as is in ws-security-server tests. I have to register custom algorithm in this test. My code does not seem like a user preferred solution. Is it possible to achieve the same via some nicer configuration?

ffang: Hi @JiriOndrusek ,

I think this is OK for now.

Freeman

ffang (approved):

integration-tests/ws-security-server/src/test/java/io/quarkiverse/cxf/it/ws/securitypolicy/server/CxfWssSecurityPolicyServerTest.java
JiriOndrusek: Hi @ffang 

I debugged this part of the test and there is a difference between natie and JVM mode. I'm setting the customAlgSuite via  `quarkus.cxf.customizedAlgorithmSuite.enabled = true`
- in JVM - bus created via QuarkusBusFactory is customized and the plain client get the same bus -> works without this line
- in native - bus created via QuarkusBusFactory is customized, but plain client is creating its own bus, which is not affected by the QuarkusBusFactory. My understanding is that the client is created by in the test (non-native) part of executions. Therefore I had to customize the client's bus to make scenario work.

I think that in JVM code works without this configuration, because both plain client (test resource) and app part (from the `main` source folder) share the customAlgSuite registered into a static variable

ppalaga: @JiriOndrusek, this kind of differences is caused by the fact that in native mode, the tests run in a JVM, but the application under test is a fully different (native) process. Hence of you register something on the Bus in the application, it is not visible for the JVM running the tests. 

In JVM mode, the tests and the application under test are a part of the same JVM.

So you either have to configure the stuff twice, or you have to move the SOAP client to the application and wrap it in REST. I think the second option is better, because it allows to cover also the client in native.

JiriOndrusek: I agree with the cause of the issue. I'll move the client creation into the application, where the customization configuration should take care of it. I'll fix it in the same way on the other places to have a consistent approach in all tests.

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -445,4 +445,116 @@ default void validate(String prefix) {
         }
     }
 
+    /**
```
ppalaga, line 448: Could we please move this to a new class in the security extension?

I'd vote for it being a Runtime fixed setting (should go to Security runtime module), so that users can query the values at runtime from their apps. 

Something like 

```
@ConfigMapping(prefix = "quarkus.cxf")
@ConfigRoot(phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
@ConfigDocFilename("quarkus-cxf-rt-ws-security.adoc")
public interface CxfSecurityFixedConfig {
...
```

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -445,4 +445,116 @@ default void validate(String prefix) {
...
+     * Parameters for fully customizable algorithm suite.
+     */
+    @WithName("customizedAlgorithmSuite")
+    public CustAlgSuite custAlgSuite();
+
+    @ConfigGroup
+    public interface CustAlgSuite {
+
+        public static final String CUSTOMIZED_ALGORITHM_SUITE_NAME = "CustomizedAlgorithmSuite";
```
ppalaga, line 457:
```suggestion
        public static final String CUSTOMIZED_ALGORITHM_SUITE_NAME = "QuarkusCxfFipsAlgorithmSuite";
```

... so that users have a chance to see where it comes from when they see it in policy files or in some sort of a debug output.

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -445,4 +445,116 @@ default void validate(String prefix) {
         }
     }
 
+    /**
+     * Parameters for fully customizable algorithm suite.
+     */
+    @WithName("customizedAlgorithmSuite")
```
ppalaga, line 451: So this whole alg suite config should be used mostly for fips compliance, right? Then I'd be for the following structure of the configuration:

```
quarkus.cxf.security.fips.enabled = [true/false] # enables the custome alg. suite
quarkus.cxf.security.fips.algorithm-suite.digest-algorithm = ...
quarkus.cxf.security.fips.algorithm-suite.encryption-algorithm = ...
...
```

WDYT?

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -445,4 +445,116 @@ default void validate(String prefix) {
...
+        public boolean enabled();
+
+        /**
+         * Digest Algorithm.
+         */
+        @WithDefault("http://www.w3.org/2001/04/xmlenc#sha256")
+        public String digestAlgorithm();
+
+        /**
+         * Encryption Algorithm.
```
ppalaga, line 495: Some hint about the possibly values would be nice, e.g. where they are defined?

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -445,4 +445,116 @@ default void validate(String prefix) {
...
+         */
+        @WithDefault("http://www.w3.org/2009/xmlenc11#aes256-gcm")
+        public String encryptionAlgorithm();
+
+        /**
+         * Symmetric Key Encryption Algorithm.
+         */
+        @WithDefault("http://www.w3.org/2001/04/xmlenc#kw-aes256")
```
ppalaga, line 503: Some hint about the possibly values would be nice, e.g. where they are defined?

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -445,4 +445,116 @@ default void validate(String prefix) {
...
+        public String symmetricKeyEncryptionAlgorithm();
+
+        /**
+         * Asymmetric Key Encryption Algorithm.
```
ppalaga, line 507: Some hint about the possibly values would be nice, e.g. where they are defined?

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -445,4 +445,116 @@ default void validate(String prefix) {
...
+         */
+        @WithDefault("http://www.w3.org/2001/04/xmlenc#rsa-1_5")
+        public String asymmetricKeyEncryptionAlgorithm();
+
+        /**
+         * Encryption Key Derivation.
```
ppalaga, line 513: Some hint about the possibly values would be nice, e.g. where they are defined?

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -445,4 +445,116 @@ default void validate(String prefix) {
...
+         */
+        @WithDefault("http://schemas.xmlsoap.org/ws/2005/02/sc/dk/p_sha1")
+        public String encryptionKeyDerivation();
+
+        /**
+         * Signature Key Derivation.
```
ppalaga, line 519: Some hint about the possibly values would be nice, e.g. where they are defined?

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -445,4 +445,116 @@ default void validate(String prefix) {
...
+         */
+        @WithDefault("http://schemas.xmlsoap.org/ws/2005/02/sc/dk/p_sha1")
+        public String signatureKeyDerivation();
+
+        /**
+         * Encryption Derived Key Length.
```
ppalaga, line 525:
```suggestion
         * Encryption Derived Key Length in <what units?>.
```

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -445,4 +445,116 @@ default void validate(String prefix) {
...
+         */
+        @WithDefault("256")
+        public Integer encryptionDerivedKeyLength();
+
+        /**
+         * Signature Derived Key Length.
```
ppalaga, line 531:
```suggestion
         * Signature Derived Key Length in <what units?>
```

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -445,4 +445,116 @@ default void validate(String prefix) {
...
+         */
+        @WithDefault("192")
+        public Integer signatureDerivedKeyLength();
+
+        /**
+         * Minimum Symmetric Key Length.
```
ppalaga, line 537:
```suggestion
         * Minimum Symmetric Key Length in <what units?>
```

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -445,4 +445,116 @@ default void validate(String prefix) {
...
+         */
+        @WithDefault("256")
+        public Integer minimumSymmetricKeyLength();
+
+        /**
+         * Maximum Symmetric Key Length.
```
ppalaga, line 543:
```suggestion
         * Maximum Symmetric Key Length in <what units?>
```

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -445,4 +445,116 @@ default void validate(String prefix) {
...
+         */
+        @WithDefault("256")
+        public Integer maximumSymmetricKeyLength();
+
+        /**
+         * Minimum Symmetric Key Length.
```
ppalaga, line 549:
```suggestion
         * Minimum Symmetric Key Length in <what units?>
```

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -445,4 +445,116 @@ default void validate(String prefix) {
...
+         */
+        @WithDefault("1024")
+        public Integer minimumAsymmetricKeyLength();
+
+        /**
+         * Maximum Symmetric Key Length.
```
ppalaga, line 555:
```suggestion
         * Maximum Symmetric Key Length in <what units?>
```

ppalaga (changes requested):
A quick review from a train. Will cont in a couple of hours

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
JiriOndrusek: I originally created it with fips property, but then I asked @ffang about his opinion. I might have understood or asked wrongly.

The advice was:
```
I'd use some name like customerizedPolicy and make all parameters configurable, it should be a general solution, so let the end users control everything
```

JiriOndrusek:
@ppalaga  thanks for the review, I'll fix all items.

I understand that using Fips in the name of the customAlgSuite would make sense. On the other hand, if user changes attributes for its own purpose, so it won't work on fips (and there will be no FIPS in user's environment), the word FIPS would be misleading

JiriOndrusek:
What about the name `QuarkusCxfCustomAlgorithmSuite` (instead of suggested QuarkusCxfFipsAlgorithmSuite)?

ppalaga:
Those are all valid proposals, however, after studying the PR in more depth, I am not sure it is a good idea to make it like this. Here is why: 

The set of AlgSuites is defined in version 1.2 of the XML Schema for the policy files: http://docs.oasis-open.org/ws-sx/ws-securitypolicy/200702/ws-securitypolicy-1.2.xsd If you CTRL-F for "7.1 AlgorithmSuite Assertion", you'll see all suites defined in https://github.com/apache/ws-wss4j/blob/master/policy/src/main/java/org/apache/wss4j/policy/model/AlgorithmSuite.java#L41-L200

The policies are used on both server and client side to configure the security and those standardized AlgSuites guarrantee that clients and servers will encrypt/sign the messages in an expected, mutually comprehensible way, across all vendors. 

Introducing an AlgSuite that is configurable through an intransparent config of the client or server would break that interoperability. Actually only Quarkus CXF clients and servers could communicate and only those ones whose custom-algorithm-suite is configured in the same way.

Saying that introducing any new AlgSuite would make the policy files non-compliant with the schema would be true, but there is already a precedent that CXF introduced new AlgSuites in response to a CVE: https://coheigea.blogspot.com/2012/04/note-on-cve-2011-1096.html I wonder wheter we should not try the same: Introduce one or more FIPS-compliant AlgSuites to https://github.com/apache/cxf/blob/main/rt/ws/security/src/main/java/org/apache/cxf/ws/security/policy/custom/DefaultAlgorithmSuiteLoader.java#L74-L114 In that way, the new suites would be at least comprehensible for clients and servers based on CXF.

WDYT?

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
ffang: Yes, I'm thinking of more general than just FIPS context. The underlying specification may stop evolving,  but if users can configure all parameters of the AlgorithmSuite to avoid using some cracked/out-dated security algo, this would be awesome. And this flexibility could not only fit FIPS scenairo.

ffang:
> Those are all valid proposals, however, after studying the PR in more depth, I am not sure it is a good idea to make it like this. Here is why:
> 
> The set of AlgSuites is defined in version 1.2 of the XML Schema for the policy files: http://docs.oasis-open.org/ws-sx/ws-securitypolicy/200702/ws-securitypolicy-1.2.xsd If you CTRL-F for "7.1 AlgorithmSuite Assertion", you'll see all suites defined in https://github.com/apache/ws-wss4j/blob/master/policy/src/main/java/org/apache/wss4j/policy/model/AlgorithmSuite.java#L41-L200
> 
> The policies are used on both server and client side to configure the security and those standardized AlgSuites guarrantee that clients and servers will encrypt/sign the messages in an expected, mutually comprehensible way, across all vendors.
> 
> Introducing an AlgSuite that is configurable through an intransparent config of the client or server would break that interoperability. Actually only Quarkus CXF clients and servers could communicate and only those ones whose custom-algorithm-suite is configured in the same way.
> 
> Saying that introducing any new AlgSuite would make the policy files non-compliant with the schema would be true, but there is already a precedent that CXF introduced new AlgSuites in response to a CVE: https://coheigea.blogspot.com/2012/04/note-on-cve-2011-1096.html I wonder wheter we should not try the same: Introduce one or more FIPS-compliant AlgSuites to https://github.com/apache/cxf/blob/main/rt/ws/security/src/main/java/org/apache/cxf/ws/security/policy/custom/DefaultAlgorithmSuiteLoader.java#L74-L114 In that way, the new suites would be at least comprehensible for clients and servers based on CXF.
> 
> WDYT?

Hi @ppalaga ,

This is a good proposal and I'm actually thinking about the same thing.

Created https://issues.apache.org/jira/browse/CXF-8971 to track this enhancement.

@JiriOndrusek , please feel free to send a PR there!

But probably this PR for quarkus-cxf still stands as don't need to wait for CXF side fix and release.

Thanks!
Freeman

JiriOndrusek:
Hi @ppalaga ,

thanks for the proposal. 
To clarify some details. Suggested solution is to move "customization" mechanism into cxf. 
If so, we can keep alg suite config parameters (after refactor to fixed ones) to configure such  algSuirta via cxf parameters.

@ffang I try to prepare a draft for cxf to fix issue  https://issues.apache.org/jira/browse/CXF-8971.

ffang:
> As I'm thinking about the customizationAlgSuite in Cxf, what about addition of a FIPS approved algSuite among https://github.com/apache/cxf/blob/main/rt/ws/security/src/main/java/org/apache/cxf/ws/security/policy/custom/DefaultAlgorithmSuiteLoader.java#L74-L114
> 
> That would make usage of the suite much easier for the users. I can imagine, that in the future there can servers customized with ALG1 or ALG2 (both works for FIPS). It might be inconvenient for the clients (because every server requires different custom alg). If there is a default fips suite, the usage can be much easier (the customization would still exist, but might not be needed in so many cases)

Hi Jiri,

Yes, the current CXF DefaultAlgorithmSuiteLoader is a good place. You can add a new policy with name "CustomerizedAlgorithmSuite" which has default FIPS compliant paramenters so it can work OOTB with zero configuration, but these parameters should be configurable/overridable from CXF bus properties. I believe you can define new added necessary properties name in org.apache.cxf.ws.security.SecurityConstants class with prefix like "ws-security.policy".

Cheers
Freeman

ppalaga:
> Created https://issues.apache.org/jira/browse/CXF-8971 to track this enhancement.

I commented [there](https://issues.apache.org/jira/browse/CXF-8971?focusedCommentId=17808768&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17808768)

JiriOndrusek:
PR for CXF is prepared (https://github.com/apache/cxf/pull/1660). I plan to refactor this PR to leverage this change (only minimum of this PR will stay as is)

JiriOndrusek:
This PR has to be reworked, because the customizableAlgSuite is proposed on the CXF source instead here. See https://github.com/apache/cxf/pull/1660

ppalaga:
@JiriOndrusek this PR builds against the staging repo of CXF 4.0.4 that is under vote right now. 
Would you like to rebase this PR on top of it? 
I'd like to release Quarkus CXF 3.8.1 on Friday once the CXF 4.0.4 artifacts are available on Central.

JiriOndrusek:
> @JiriOndrusek [this PR](https://github.com/quarkiverse/quarkus-cxf/pull/1274) builds against the staging repo of CXF 4.0.4 that is under vote right now. Would you like to rebase this PR on top of it? I'd like to release Quarkus CXF 3.8.1 on Friday once the CXF 4.0.4 artifacts are available on Central.

Thanks for he upgrading PR, I'll look at that and rebase this PR.

JiriOndrusek:
@ppalaga @ffang Is the staging repo working now? I can not download any artifact from it and the url does not work in the browser: https://repository.apache.org/content/repositories/orgapachecxf-1196

JiriOndrusek:
Shouldn't it be https://repository.apache.org/content/repositories/orgapachecxf-1197/ ?

JiriOndrusek:
I switch staging repo to https://repository.apache.org/content/repositories/orgapachecxf-1197/ to make the CI pass (new repo is part of the second voting)

ppalaga:
Thanks for rabasing, @JiriOndrusek !
I took over you commit to https://github.com/quarkiverse/quarkus-cxf/pull/1274
and added some fixups https://github.com/quarkiverse/quarkus-cxf/pull/1274/commits/78e12a8346118f1a3b9f6b24cd70c5ee46bc78f8 - could you please review there?
I'd also add a bit more docs text to the new options.

ppalaga:
Replaced by https://github.com/quarkiverse/quarkus-cxf/pull/1274

PR #1237, opened 2024-02-15
[Docs] Fix ws-rm server link after app split

avano:
@ppalaga I fixed only the link to point to the server-jvm, as the native doesn't contain the actual code. if you prefer, I can also add a link to the native app


ppalaga:
Indeed, thanks for spotting it, @avano! I unfortunately cannot merge this, because the file you fixed is generated from [this one](https://github.com/quarkiverse/quarkus-cxf/blob/main/extensions/ws-rm/runtime/src/main/doc/usage.adoc?plain=1#L18) I fixed it here https://github.com/quarkiverse/quarkus-cxf/commit/e1ff89a292cd8fcccf1bb5c555a84a281346b03f thanks again!

avano:
oops, sorry :sweat_smile:

PR #1274, opened 2024-03-05
Upgrade to CXF 4.0.4

ppalaga:


JiriOndrusek:
@ppalaga thank you for `Fixup make algorithm suites configurable through application.properties`
(Mainly missing native test class)
I agree wit those changes.

Issue #1285, opened 2024-03-12
Move tests passing on a FIPS system from ws-security-policy to ws-security-policy-fips


ppalaga:
There is perhaps more tests in https://github.com/quarkiverse/quarkus-cxf/tree/main/integration-tests/ws-security-policy that would pass on a system with FIPS assertions enabled. We should move those to https://github.com/quarkiverse/quarkus-cxf/tree/main/integration-tests/ws-security-policy-fips

PR #1286, opened 2024-03-12
Move tests passing on a FIPS system

JiriOndrusek:
fixes https://github.com/quarkiverse/quarkus-cxf/issues/1285

1. Class `TransportPolicyTest` moved from `ws-security-policy` to `ws-security-policy-fips`.
2. Test `CustomEncryptSignPolicyTest` was modified to be able to detect FIPS machine and change assertions accordingly.
3. Added missing `EncryptSignPolicyIT`.

This solution duplicates several java classes: 
- `HelloService` 
- `HelloServiceImpl`
- part of `BeanProducers`
- part of `PolicyTestUtils`

I can imagine a different solution avoiding duplicities:
- The number of tests failing in FIPS is quite small (only 5 tests are failing). It would be perhaps clearer if those 5 methods would be handled in the same way as `CustomEncryptSignPolicyTest` to change assertions in case a FIPS machine is present. 

This second solution would:
- get rid of separated module
- get rid of duplicities
- no blacklisting of `ws-security-policy` would be required anymore.


JiriOndrusek:
@ppalaga WDYT ^ (second solution)

integration-tests/ws-security-policy-fips/src/test/java/io/quarkiverse/cxf/it/security/policy/PolicyTestUtils.java
```diff
@@ -14,4 +21,39 @@ public static RestAssuredConfig restAssuredConfig() {
                 "client-truststore-password"));
     }
 
+    public static boolean isFipsEnabled() throws IOException {
+        return java.security.Security.getProvider("SunPKCS11-NSS-FIPS") != null;
+
+        //rhel specific
+        //        return "1".equals(Files.readString(Paths.get("/proc/sys/crypto/fips_enabled"), StandardCharsets.UTF_8));
```
JiriOndrusek, line 28: I think that solution with security provider is more general.

On the other hand, the rhel specific solution is quite faster.

I'm inclined to use a provider for detection, but I can change the solution, if anybody prefers different one,

ppalaga:
> The number of tests failing in FIPS is quite small (only 5 tests are failing). It would be perhaps clearer if those 5 methods would be handled in the same way as CustomEncryptSignPolicyTest to change assertions in case a FIPS machine is present.

Yeah, what you say makes a lot of sense. Sorry for my suggestions in the other direction. I thought it was much more tests to handle. Would you please implement that?

JiriOndrusek:
> Yeah, what you say makes a lot of sense. Sorry for my suggestions in the other direction. I thought it was much more tests to handle. Would you please implement that?

I started working on the change, but I found a weird behavior witch originates in cxf (I think). I reported the bug as https://issues.apache.org/jira/browse/CXF-8986 

I pinged @ffang in the cxf issue. I will finish this PR after confirmation whether the behavior is a bug or a feature.

JiriOndrusek:
Superseded by https://github.com/quarkiverse/quarkus-cxf/pull/1297

PR #1308, opened 2024-03-21
Fixes for testing on Win

avano:
- use the actual container host in client test (when running the container on a RHEL host using `DOCKER_HOST` variable)
- execute `generate-certs.sh` script using `bash` (should be available for both cygwin and WSL)


ppalaga:
Thanks @avano!

PR #1328, opened 2024-04-02
XmlSchemaValidationTest fails on the different language settings

### LINKED ISSUE #1327: XmlSchemaValidationTest fails on the different language settings
I'm running the tests on my local machine which is using `zh_CN.utf-8`. Then `XmlSchemaValidationTest` is failing with

```
[ERROR]   XmlSchemaValidationTest.addCheckResultInvalid:74 
Expecting throwable message:
  "Unmarshalling Error: cvc-maxInclusive-valid: 对于类型为 'smallIntType' 的 maxInclusive '1024', 值 '1025' 不具有面有效性。 "
to contain:
  "arshalling Error: cvc-maxInclusive-valid: Value '1025' is not facet-valid with respect to maxInclusive '1024' for type 'smallIntType'."
but did not.
```

The cause message looks like from `i18n` resource bunlde and it transforms some words to Chinese. This can be fixed by setting `LANG` environment.

ppalaga:
fix #1327

### DIFF
--- a/docs/modules/ROOT/examples/client-server/application.properties
+++ b/docs/modules/ROOT/examples/client-server/application.properties
@@ -81,3 +81,5 @@ quarkus.cxf.client.basicAuthSecureWsdl.username = bob
 quarkus.cxf.client.basicAuthSecureWsdl.password = ${client-server.bob.password}
 quarkus.cxf.client.basicAuthSecureWsdl.secure-wsdl-access = true
 # end::client-basic-auth-wsdl[]
+
+quarkus.default-locale = en_US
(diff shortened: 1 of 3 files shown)


shumonsharif:
@ppalaga I took a quick look here, I believe the LANG environment variable is Unix/Linux specific, and this fix won't work on Windows.  Is there a reason to avoid the Java args:  `-Duser.language=en -Duser.country=US` which should have the same effect while being cross platform?

ppalaga:
Thanks for the hint @shumonsharif , tried with those sysprops but if the env has `LANG=zh_CN.utf-8` the env wins. So I am keeping both.

Issue #1348, opened 2024-04-17
Ws-security tests: fails in fips because of missing  <sp:CustomAlgorithmSuite/>


JiriOndrusek:
It seems, that one place was skipped in https://github.com/quarkiverse/quarkus-cxf/commit/96405020012ec2c95e537dd7392df4e067e1ca95

The missing place is https://github.com/quarkiverse/quarkus-cxf/blob/main/integration-tests/ws-security/src/main/resources/server/encrypt-sign-policy.xml#L30

PR #1349, opened 2024-04-17
Fixed alg suite for FIPS environment in ws-security tests

JiriOndrusek:
fixes https://github.com/quarkiverse/quarkus-cxf/issues/1348


docs/modules/ROOT/examples/calculator-client/pom.xml
```diff
@@ -4,7 +4,7 @@
     <parent>
         <groupId>io.quarkiverse.cxf</groupId>
         <artifactId>quarkus-cxf-integration-tests</artifactId>
-        <version>3.9.1-SNAPSHOT</version>
+        <version>3.10.0-SNAPSHOT</version>
```
JiriOndrusek, line 7: @ppalaga I had to update this version, to make CI pass. I think that either:

- control of version is too strict for this example
- release script has to be changed to include this file for version upgrade

ppalaga: Yeah, it's pretty stupid. I have pushed a "Fixup release" commit manually in the past. Now I have implemented a proper fix in  https://github.com/quarkiverse/quarkus-cxf/commit/23ff3436f12f67f1459a87b9d82e111a65c6c05b

ppalaga (commented):
I force-pushed to your branch without the pom.xml change.

ppalaga (approved):

PR #1499, opened 2024-09-09
Avoid publishing support modules in Maven Central

gastaldi:
- As requested in https://github.com/quarkiverse/quarkiverse-devops/issues/269

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -111,10 +111,8 @@
     <modules>
         <module>bom</module>
         <module>extensions</module>
-        <module>docs</module>
         <module>bom-test</module>
         <module>test-util-parent</module>
-        <module>integration-tests</module>
     </modules>
 
     <build>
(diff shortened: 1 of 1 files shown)


gastaldi:
> Quarkus CXF is a Platform member, due to which the integration test modules incl. test jars are needed for testing the Platform.
> 
Ok, I updated the PR to keep it.

> Were those Central rules introduced recently?
That's explained in https://central.sonatype.org/publish/requirements/#supply-javadoc-and-sources and JReleaser (the tool we'll use to push to Central) follows it rigorously to be able to publish to Central 

> 
> Our docs module actually has a bytecode jar, source jar, because there is application.properties. The only missing part is JavaDoc, which could be added by adding a fake Java source file.

The docs module isn't referenced anywhere, and I believe adding a fake Java source file just so it's published in central is pointless.

ppalaga (approved):

PR #1536, opened 2024-10-03
Per client or service configuration of sensitive-element-names and sensitive-protocol-header-names not honored 

### LINKED ISSUE #1533: Per client or service configuration of `sensitive-element-names` and `sensitive-protocol-header-names` not honored
This week I discovered that our client specific configuration for `sensitive-element-names` stopped working after upgrading from 2.5.0 to 3.8.2 all the way back in May, i.e. config properties of the form `quarkus.cxf.client."client-name".logging.sensitive-protocol-header-names`

I did some incremental updates and when updating from 2.6.1 to 2.7.0.CR1 the configuration stopped working.

Comparing: https://github.com/quarkiverse/quarkus-cxf/compare/2.6.1...2.7.0.CR1 it seems that the structure of lines 121-122 of the old file was copied in `LoggingFactoryCustomizer.java` lines [78-79](https://github.com/quarkiverse/quarkus-cxf/blob/main/extensions/core/runtime/src/main/java/io/quarkiverse/cxf/logging/LoggingFactoryCustomizer.java#L78-L79) without taking into account any configuration present in the `PerClientOrServiceLoggingConfig`  object and so that only the global config is added to the configured LoggingFeature.

As per client logging configuration is still documented I assume this must be a bug.

ppalaga:
Fix #1533

### DIFF
--- a/extensions/core/deployment/src/test/java/io/quarkiverse/cxf/deployment/logging/GlobalSensitiveTest.java
+++ b/extensions/core/deployment/src/test/java/io/quarkiverse/cxf/deployment/logging/GlobalSensitiveTest.java
@@ -1,0 +1,259 @@
+package io.quarkiverse.cxf.deployment.logging;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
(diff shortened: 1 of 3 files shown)


ppalaga:
Ups, I merged by mistake.

PR #1587, opened 2024-11-08
Remove findbugs exclusion since it is needed by quarkus-grpc-common

zhfeng:


ppalaga:
@zhfeng could you please explain, what problem would this change solve? There is no dependency on GRPC in Quarkus CXF.

zhfeng:
Well, it is a transitive dep via `io.quarkus:quarkus-opentelemetry`. I suppose there might be a potential issue like 
- https://github.com/apache/camel-quarkus/issues/5167
- https://github.com/apache/camel-quarkus/issues/6754

ppalaga:
OK, thanks, I see that `io.quarkus:quarkus-opentelemetry` actually depends on `quarkus-grpc-common`

PR #1610, opened 2024-11-20
Make the CI more modular to allow for testing also against CXF and branches other than main

ppalaga:


ppalaga:
Replaced by https://github.com/quarkiverse/quarkus-cxf/pull/1768

PR #1669, opened 2024-12-20
Add quarkus.cxf.client."client-name".max-same-uri configuration option 

### LINKED ISSUE #1639: Add `quarkus.cxf.client."client-name".max-same-uri` configuration option
The default implementation of the HTTP Conduit now uses the VertxHttpClientHTTPConduit, which throws an Redirection Loop error when redirected to the same URL if the http.redirect.max.same.uri.count property is not set. There is no quarkus property corresponding to this property and it needs to be set as a system property when starting the app. 

This is in contrast to the original default (in quarkus 3.15 and before) URLConnectionHTTPConduit where redirection (also to the same URL) are automatically enabled so no properties need to be set. 

Would it be possible to add a quarkus property corresponding to this for easier usage? Thank you!

ppalaga:
Fix #1639

@dcheng1248 I dared to pick your work from https://github.com/dcheng1248/quarkus-cxf/pull/1 and adapted to make it work the way I think is appropriate. Please review, if you like.

### DIFF
--- a/extensions/core/runtime/src/main/java/io/quarkiverse/cxf/vertx/http/client/VertxHttpClientHTTPConduit.java
+++ b/extensions/core/runtime/src/main/java/io/quarkiverse/cxf/vertx/http/client/VertxHttpClientHTTPConduit.java
@@ -25,22 +25,10 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.PushbackInputStream;
-import java.net.HttpURLConnection;
-import java.net.InetSocketAddress;
-import java.net.Proxy;
+import java.net.*;
 import java.net.Proxy.Type;
-import java.net.SocketTimeoutException;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.util.ArrayDeque;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.Map.Entry;
-import java.util.Queue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
@@ -61,12 +49,7 @@ import org.apache.cxf.phase.PhaseInterceptorChain;
 import org.apache.cxf.service.model.EndpointInfo;
 import org.apache.cxf.transport.Conduit;
 import org.apache.cxf.transport.MessageObserver;
-import org.apache.cxf.transport.http.Address;
-import org.apache.cxf.transport.http.Cookies;
-import org.apache.cxf.transport.http.HTTPConduit;
-import org.apache.cxf.transport.http.HTTPException;
-import org.apache.cxf.transport.http.Headers;
-import org.apache.cxf.transport.http.MessageTrustDecider;
+import org.apache.cxf.transport.http.*;
 import org.apache.cxf.transports.http.configuration.HTTPClientPolicy;
 import org.apache.cxf.version.Version;
 import org.apache.cxf.ws.addressing.EndpointReferenceType;
@@ -80,19 +63,9 @@ import io.quarkus.arc.Arc;
 import io.quarkus.arc.InstanceHandle;
 import io.quarkus.logging.Log;
 import io.quarkus.runtime.BlockingOperationControl;
-import io.vertx.core.AsyncResult;
-import io.vertx.core.Future;
-import io.vertx.core.Handler;
-import io.vertx.core.MultiMap;
-import io.vertx.core.Promise;
+import io.vertx.core.*;
 import io.vertx.core.buffer.Buffer;
-import io.vertx.core.http.HttpClient;
-import io.vertx.core.http.HttpClientRequest;
-import io.vertx.core.http.HttpClientResponse;
-import io.vertx.core.http.HttpHeaders;
-import io.vertx.core.http.HttpMethod;
-import io.vertx.core.http.HttpVersion;
-import io.vertx.core.http.RequestOptions;
+import io.vertx.core.http.*;
 import io.vertx.core.http.impl.HttpUtils;
 import io.vertx.core.net.ProxyOptions;
 import io.vertx.core.net.ProxyType;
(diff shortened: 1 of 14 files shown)


dcheng1248:
> Fix #1639
> 
> @dcheng1248 I dared to pick your work from [dcheng1248#1](https://github.com/dcheng1248/quarkus-cxf/pull/1) and adapted to make it work the way I think is appropriate. Please review, if you like.

Will do! Thanks for picking this up and sorry for the delay. I had a high fever the past two days so everything paused - but back on now. When do you plan to have this merged?

integration-tests/client-server/src/main/java/io/quarkiverse/cxf/it/redirect/RedirectRest.java
```diff
@@ -93,16 +117,25 @@ public Response tripleRedirect() {
...
-    public Response selfRedirect(@Context HttpHeaders headers) {
+    public Response selfRedirect(
+            @PathParam("selfRedirectsCount") int selfRedirectsCount,
+            @PathParam("resetCount") int resetCount) {
         int count = redirectCounter.incrementAndGet();
-        if (count <= NUM_SELF_REDIRECTS) {
-            return Response.status(302).header("Location", "/RedirectRest/selfRedirect").build();
+
+        if (selfRedirectsCount == resetCount && count == resetCount) {
+            redirectCounter.set(0);
```
dcheng1248, line 128: This is a smart way to do the counting, but one remaining issue is it doesn't really reset the counter to 0 after a test for too many same URIs. You can see that with the current setup with async/maxSameUri2, the counter begins at 1 instead of 0 due to the previous sync test. I don't have a good solution for this (other than to add yet another path param that copies the max same uri property), but would perhaps set the counter to -1 instead of 0 here to at least avoid this problem in this setup in case future tests are added.

integration-tests/client-server/src/test/java/io/quarkiverse/cxf/it/redirect/RedirectTest.java
```diff
@@ -70,7 +86,15 @@ void doubleRedirectMaxRetransmits1(String endpoint) {
...
-                .body(CoreMatchers.containsString("Unexpected EOF in prolog"));
+                .body(
+                        CoreMatchers.either(
+                                Matchers.matchesPattern("\\QReceived redirection status 307 from"
+                                        + " http://localhost:\\E[0-9]+\\Q/RedirectRest/singleRedirect"
+                                        + " by client doubleRedirectMaxRetransmits1,"
+                                        + " but already performed maximum number 1 of allowed retransmits;"
+                                        + " you may want to increase quarkus.cxf.client.doubleRedirectMaxRetransmits1.max-retransmits."
+                                        + " Visited URIs: http://localhost:\\E[0-9]+\\Q/RedirectRest/doubleRedirect -> http://localhost:\\E[0-9]+\\Q/RedirectRest/singleRedirect\\E"))
+                                .or(CoreMatchers.containsString("Unexpected EOF in prolog")));
```
dcheng1248, line 97: question: why is the second matcher necessary? The test seems to only return the first string. Same for the other tests with two matchers.

extensions/core/runtime/src/main/java/io/quarkiverse/cxf/vertx/http/client/VertxHttpClientHTTPConduit.java
```diff
@@ -657,11 +694,15 @@ void finishRequest(HttpClientRequest req, Buffer buffer) {
 
         }
 
+        private static int performedRetransmits(List<URI> retransmits) {
+            /* The first element in the retransmits list is the original URI that we do not count as a retransmit */
+            return retransmits.size() - 1;
+        }
+
         void redirectRetransmit(URI newURL) throws IOException {
             if (Log.isDebugEnabled()) {
```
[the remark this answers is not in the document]
ppalaga: The if statement could indeed be omitted, if the log statement params would all be some simple values. But in this case, there is `redirects.stream().map(URI::toString).collect(Collectors.joining(" -> "))` which costs some CPU cycles, but is useless in case the debug level is not active.

integration-tests/client-server/src/main/java/io/quarkiverse/cxf/it/redirect/RedirectRest.java
ppalaga: Oh, really, thanks for finding out. We could introduce a separate rest endpoint to reset the counter.

integration-tests/client-server/src/test/java/io/quarkiverse/cxf/it/redirect/RedirectTest.java
ppalaga: The same tests are run with `QUARKUS_CXF_DEFAULT_HTTP_CONDUIT_FACTORY=URLConnectionHTTPConduitFactory` environment variable https://github.com/quarkiverse/quarkus-cxf/blob/main/.github/actions/build-and-run-jvm-tests/action.yml#L44-L49
The alternative set of error messages is for the URLConnectionHTTPConduit.

extensions/core/runtime/src/main/java/io/quarkiverse/cxf/vertx/http/client/VertxHttpClientHTTPConduit.java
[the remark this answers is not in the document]
dcheng1248: I never thought about it from a performance POV, but that makes a lot of sense, thanks!

integration-tests/client-server/src/main/java/io/quarkiverse/cxf/it/redirect/RedirectRest.java
dcheng1248: Do you mean a separate endpoint only for the too many same uri test, which basically doesn't keep a counter and just keeps redirecting to self? So this endpoint is only kept for the tests with max-uri > number of self redirects, which means it always resets to 0 after?

integration-tests/client-server/src/test/java/io/quarkiverse/cxf/it/redirect/RedirectTest.java
dcheng1248: Didn't realize that, thank you!

integration-tests/client-server/src/main/java/io/quarkiverse/cxf/it/redirect/RedirectRest.java
dcheng1248: I looked at it again after a night's sleep. I think a new endpoint is not needed. If you change the URI of the maxSameURi2 test to selfRedirect/3/3 it should solve the issue, which makes sense because the selfRedirectCount indicates how many self redirects there should be before the test completes (either going to a real endpoint or throwing an error), and for maxSameUri2 we need 3 self redirects before an error is thrown. 

The condition can also be simplified to 
`
if (count == resetCount) {
    redirectCounter.set(0);
}
`

ppalaga: See 7bb1c1b where I added `RedirectRest.selfRedirect()` for resetting the counter.

dcheng1248: I never thought about it that way, interesting solution. Question: why do the try-finally in the redirect test instead of splitting it into two tests (small/large body) and let the BeforeEach method do the job?

ppalaga: No specific reason, could be done that way too.

PR #1686, opened 2025-01-20
Fix Ecosystem CI workflow

gsmet:
This workflow has been broken for quite a while.
I noticed it as part of my work to detect stale Ecosystem CI builds.

### DIFF
--- a/.github/workflows/quarkus-snapshot.yaml
+++ b/.github/workflows/quarkus-snapshot.yaml
@@ -22,13 +22,14 @@ jobs:
   build:
     name: "Build against latest Quarkus snapshot"
     runs-on: ubuntu-latest
-    if: github.actor == 'quarkusbot' || github.actor == 'quarkiversebot' || github.actor == 'dufoli'
+    if: github.actor == 'quarkusbot' || github.actor == 'quarkiversebot'
 
     steps:
       - name: Set up Java
         uses: actions/setup-java@v4
         with:
           java-version: ${{ env.JAVA_VERSION }}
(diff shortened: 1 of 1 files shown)


gsmet:
Fast-tracking it to get the status tomorrow.

ppalaga:
@gsmet, could you please help to refresh my memory  where the Ecosystem CI is documented? Is it testing the platform or just latest Quarkus CXF and Quarkus core?

gsmet:
It's testing CXF main + Quarkus Core main (the way it's configured now).

You can find more information here: https://github.com/quarkusio/quarkus-ecosystem-ci/?tab=readme-ov-file#what-its-all-about .

ppalaga:
Thanks @gsmet. We tend to have our own similar workflow in Camel Quarkus and Quarkus CXF that works with `quarkus-main` branch that may contain fixes specific to the latest `main` branch of Quarkus. Our workflow rebases `quarkus-main` on top of `main` before building CQ/QCXF. I also started some works to extract parts of that workflow into a reusable GH actions to make it possible to create workflows also for main and LTS branches of other major dependencies, such as Camel and CXF. I wonder whether/how we could reconcile these efforts? Having two similar workflows is not that useful.

gsmet:
The nice thing about Ecosystem CI is that it's consistent for all Quarkiverse extensions and centralized on https://status.quarkus.io/ .

I think it has value for us (Quarkus team) to keep an eye of Quarkus CXF status, this way or another.

PR #1702, opened 2025-01-30
Run FastInfosetTest#fastInfosetNative 50 times

ppalaga:


ppalaga:
Solved in https://github.com/quarkiverse/quarkus-cxf/pull/1701

PR #1721, opened 2025-02-14
Introduce quarkus.cxf.client.tls-configuration-name property

dcheng1248:
Created as draft for code comments and review, and because it should be rebased with temp functionalities removed after we move to Quarkus 3.19.0.CR1.

I have only run the JVM tests. I have not added the relevant methods and fields for native registration because they are not needed after the rebase. 

A few things I am not quite satisfied with, but haven't come up with a better solution and am hoping for suggestions:
1. Documentation - are the current updates enough? Should we provide examples?
2. Unit test - ideally, I would like to use different truststore files for the different test cases (JVM default, Quarkus default, Named), but I haven't been able to find a setting that allows multiple keystores. I have tested each case in isolation locally and they work.
3. Unit test - I have not tested the case where the `javax.net.ssl.trustStore` or `javax.net.ssl.trustStorePassword` system properties are changed. I'm not sure if it's even possible to test that without creating a different maven profile? 
4. Integration test - I did not see an obvious place to do an integration test. The closest would be the `TransportPolicyTest` but it also isn't quite fitting. Do you think we need to add an IT? 

Thanks!


extensions/core/deployment/src/test/java/io/quarkiverse/cxf/deployment/test/TlsConfigurationTest.java
```diff
@@ -44,69 +52,247 @@ public class TlsConfigurationTest {
...
+
+    @CXFClient("helloUrlConnection3")
+    HelloService helloUrlConnection3;
+
     @Inject
     Logger logger;
 
+    @PostConstruct
+    void setup() throws Exception {
+        addCertToDefaultCacert();
```
ppalaga, line 194: I'd vote for not trying to change any file outside of the target folder of the current Maven module or temp folder. If Java is installed system-wide then the cacert file might not be writable at all. Otherwise I doubt anybody would be happy to have the default truststore of the current JDK changed by some random test.

Can't we copy the cacert file to target folder, enhance the copy and then pass it via `javax.net.ssl.trustStore` sysprop to the test?

ppalaga:
Thanks for the contribution, @dcheng1248! 

The `main` branch ist on Quarkus 3.19.0.CR1 now. Could you please rebase?

> * Documentation - are the current updates enough? Should we provide examples?

I'd review that after you rebase. 

> * Unit test - ideally, I would like to use different truststore files for the different test cases (JVM default, Quarkus default, Named), but I haven't been able to find a setting that allows multiple keystores. I have tested each case in isolation locally and they work.

The `@Certificates` annotation allows for creating multiple trustores. Check for an example in `CertReloadTest`.

> * Unit test - I have not tested the case where the `javax.net.ssl.trustStore` or `javax.net.ssl.trustStorePassword` system properties are changed. I'm not sure if it's even possible to test that without creating a different maven profile?

See below

> * Integration test - I did not see an obvious place to do an integration test. The closest would be the `TransportPolicyTest` but it also isn't quite fitting. Do you think we need to add an IT?

Integration tests are important to see the stuff working in native mode. Even playing with the copy of cacerts file would be a bit easier within an integration test. Implementing a `QuarkusTestResourceLifecycleManager` allows you to execute some code before starting the test application and passing some system properties to it. `QuarkusTestResourceLifecycleManager` is a good example. 

You may want to add your tests to the `quarkus-cxf-integration-test-client-server` module.

dcheng1248:
> The `main` branch ist on Quarkus 3.19.0.CR1 now. Could you please rebase?

I rebased, but the `Duration.parse("10S")` exception on 3.19.0.CR1 is causing the VertxJVMDefault test to fail. Would it be possible to fix and backport this in Quarkus? 

I will work on the other comments soon.

extensions/core/deployment/src/test/java/io/quarkiverse/cxf/deployment/test/TlsConfigurationTest.java
dcheng1248: I have updated and set up the unit tests to directly use certs generated in the target folder, and will test the behaviour of the default truststore with ITs.

ppalaga:
> but the `Duration.parse("10S")` exception on 3.19.0.CR1 is causing the VertxJVMDefault test to fail

Sorry about that, I have copied something without thinking. Thanks for finding it! I have sent a fix to Quarkus https://github.com/quarkusio/quarkus/pull/46310

ppalaga:
> I'm travelling soon and am only sparsely available the coming week. [...] If time is short, please feel free to edit and merge. 

Yeah, I'll try to finish this for 3.19 that I'll be releasing around Thursday Feb. 20.

PR #1723, opened 2025-02-19
Upgrade to Quarkus 3.19.0, Introduce quarkus.cxf.client.tls-configuration-name to set TLS options for all CXF clients 

### LINKED ISSUE #1680: Introduce `quarkus.cxf.client.tls-configuration-name` to set TLS options for all CXF clients
Currently, to set the TLS truststore for a client, the quarkus property `quarkus.cxf.client."client-name".tls-configuration-name` must be set. In our use case, we have all the certificates gathered into one truststore but have several clients, and it can feel cumbersome to have to set the property for every client. It would help avoid repeated code if an app-wide property is introduced.

I would be happy to see if I can help implement this if you think it is worth doing but not high priority.

ppalaga:
Fix #1680

### DIFF
--- a/docs/antora.yml
+++ b/docs/antora.yml
@@ -8,7 +8,7 @@ asciidoc:
   attributes:
 
     # Versions
-    quarkus-version: 3.19.0.CR1 # replace ${quarkus.version}
+    quarkus-version: 3.19.0 # replace ${quarkus.version}
     quarkus-cxf-version: 3.18.1 # replace ${release.current-version}
 
     # Toggle whether some page elements are rendered
--- a/docs/modules/ROOT/examples/client-server/application.properties
+++ b/docs/modules/ROOT/examples/client-server/application.properties
@@ -201,4 +201,11 @@ quarkus.cxf.client.retransmitCache.service-interface = io.quarkiverse.cxf.it.red
 quarkus.cxf.client.retransmitCache.auto-redirect = true
 #quarkus.cxf.client.retransmitCache.logging.enabled = true
 
+
+# ClientTlsTest
+# Server TLS is set in JavaNetSslClientTestResource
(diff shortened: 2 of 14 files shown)


ppalaga:
@dcheng1248 I took your work from https://github.com/quarkiverse/quarkus-cxf/pull/1721 and adapted mostly only the tests. The rest was nearly perfect. Great work, thanks a lot!

dcheng1248:
@ppalaga thanks a lot for following up and your patience throughout! The tests look much cleaner.
When time allows next week, I will try to learn more about the server side and figure out why the ITs I wrote fail. If I have trouble with this, could I ask you for quick pointers?

ppalaga:
0108d44: just rebased. Many native tests are still expected to fail due to https://github.com/quarkusio/quarkus/issues/46368 
The [fix](https://github.com/quarkusio/quarkus/commit/ababb2fe21578071d257e37cef8ba310d9ab3875) is merged in Quarkus main and will be backported to Quarkus 3.19.1 due on Wed Feb 26, 2025. 
I am going to merge with the failing native tests, release QCXF 3.19.0 and document the issue in release notes.

dcheng1248:
> There is no big mistery :D I just screwed the first attempt to implement `javax.ssl.net` TLS config on Quarkus side: [quarkusio/quarkus#46368](https://github.com/quarkusio/quarkus/issues/46368) The fix is underway: [quarkusio/quarkus#46379](https://github.com/quarkusio/quarkus/pull/46379)

My first implementation of the ITs also failed on JVM so there's something additional that's wrong - I suspect how I implemented the service, because once you used /soap/soap12 it works. For some reason even if I set the client endpoint to quarkus-tls-test (supposedly 8444) it ends up getting sent to the 8080 port (which isn't even the non-tls test port!) according to the cxf request logs. It's probably my lack of understanding of how server-side works that's causing issues - I will try to remedy this as soon as I get some time on hand.

PR #1960, opened 2025-10-26
Anchor fixes in quarkus-cxf-rt-ws-security.adoc

rhaetor:
1: The anchor [[custom-algorithm-suite-example]] is duplicated, which breaks downstream build. 

Replaced all occurences of #custom-algorithm-suite-example 
with #custom-algorithm-suite-example-endpoint or 
#custom-algorithm-suite-example-client, 
and the corresponding anchors.

2: Some anchors won't resolve automatically in the downstream, so I added manual ones, and changed referencing links

[[issue_1633_httpclienthttpconduitfactory_value_deprecated]]
[[vert-x-httpclient-based-http-conduit-default]]

### DIFF
--- a/docs/modules/ROOT/pages/release-notes/3.13.1.adoc
+++ b/docs/modules/ROOT/pages/release-notes/3.13.1.adoc
@@ -9,7 +9,7 @@
 
 === https://github.com/quarkiverse/quarkus-cxf/issues/1455[#1455] Vert.x HTTP client is incorrectly used as the default
 
-We introduced xref:release-notes/3.13.0.adoc#vert-x-httpclient-based-http-conduit-default[Vert.x HttpClient based HTTP Conduit] in {quarkus-cxf-project-name} 3.13.0
+We introduced xref:release-notes/3.13.0.adoc#vert-x-httpclient-based-http-conduit[Vert.x HttpClient based HTTP Conduit] in {quarkus-cxf-project-name} 3.13.0
 and it was not supposed to replace the original `HttpURLConnection` based default conduit.
 However, we screwed the default handling which resulted in using `VertxHttpClientHTTPConduit` as a default in some cases.
 This was fixed in {quarkus-cxf-project-name} 3.13.1.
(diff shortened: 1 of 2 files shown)


ppalaga:
> 1: The anchor [[custom-algorithm-suite-example]] is duplicated, which breaks downstream build.
> 
> Replaced all occurences of #custom-algorithm-suite-example with #custom-algorithm-suite-example-endpoint or #custom-algorithm-suite-example-client, and the corresponding anchors.

The replacement won't work like this because that anchor comes from a single configuration node type, that is used twice. 
I think the easiest fix would be to remove the `custom-algorithm-suite-example` def and make the existing refs to refer to some of the two config properties (`quarkus-cxf_quarkus-cxf-client-client-name-security-custom-digest-algorithm` or `quarkus-cxf_quarkus-cxf-endpoint-endpoint-path-security-custom-digest-algorithm`), which are already unique. Those two are just a couple of lines above the example.

please do the following: 

1. Remove `[[custom-algorithm-suite-example]]` in https://github.com/quarkiverse/quarkus-cxf/blob/main/extensions/ws-security/runtime/src/main/java/io/quarkiverse/cxf/ws/security/CxfWsSecurityConfig.java#L1066
2. Replace `#custom-algorithm-suite-example` with either `quarkus-cxf_quarkus-cxf-client-client-name-security-custom-digest-algorithm` or `quarkus-cxf_quarkus-cxf-endpoint-endpoint-path-security-custom-digest-algorithm` in https://github.com/quarkiverse/quarkus-cxf/blob/main/extensions/ws-security/runtime/src/main/java/io/quarkiverse/cxf/ws/security/CxfWsSecurityConfig.java and https://github.com/quarkiverse/quarkus-cxf/blob/main/docs/modules/ROOT/pages/release-notes/3.8.1.adoc
3. Rebuild the whole project using `./mvnw clean install -DskipTests -Dquarkus.build.skip` to re-generate the adoc files in the docs module

PR #1980, opened 2025-11-25
Expose http keepalive timeout related options to be configured

dankristensen:
Expose keep alive related timeout values, so they can be configured individually.

We are concurrently communicating with a Mainframe, that closes the connections after 10 seconds, so we need to be able to configure this on the client side, to avoid any "Connection was closed" errors.

I looked at this commit https://github.com/quarkiverse/quarkus-cxf/commit/fc0a9b3a7ffce6b8cbc7be35f30f6274df263253#diff-02109b73d940f5a3ec39ed8014c7da1e2dfe291ed458bf2a92e7785789aa3512, but could not find any tests related to this, so i have not created any. 

Please guide me if you think this is possible

### DIFF
--- a/extensions/core/deployment/src/test/java/io/quarkiverse/cxf/vertx/http/client/KeepAliveTimeoutTest.java
+++ b/extensions/core/deployment/src/test/java/io/quarkiverse/cxf/vertx/http/client/KeepAliveTimeoutTest.java
@@ -1,0 +1,88 @@
+package io.quarkiverse.cxf.vertx.http.client;
+
+import io.quarkiverse.cxf.annotation.CXFClient;
+import io.quarkus.test.QuarkusUnitTest;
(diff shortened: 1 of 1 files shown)


ppalaga:
Cool, thanks for the improvements! Let me find a similar test.

dankristensen:
I have added the test for the Connection header (close / Keep-Alive). But we cannot add a test for the timeout's, since these are passed down to vert.x, in order to tell it when to cleanup connections in it's pool. I have tried resolving the conflict, and getting it pushed correctly, without success. Do you mind looking at this? Thanx a lot for your support @ppalaga

ppalaga:
@dankristensen sorry for the delay, I got distracted by other tasks. 

I dared to squash your first two commits to https://github.com/quarkiverse/quarkus-cxf/pull/1999 and wrote the tests anew.

PR #1983, opened 2025-11-26
[Ciphersuites] Fix root cause message on windows

avano:
when running on windows 2022, the test failed as the message was slightly different:

```
Expecting a root cause with message:
  "Received fatal alert: handshake_failure"
but message was:
  "(handshake_failure) Received fatal alert: handshake_failure".
```


ppalaga (approved):
Thanks @avano

ppalaga:
I have backported to 3.27

PR #2104, opened 2026-04-19
Reduce log verbosity in tests

ppalaga:


ppalaga:
cc @vkasala

vkasala:
Thank you

PR #2117, opened 2026-04-27
Forward configurable -D properties to child Maven invocations in DevModeTest

vkasala:
Use -Dqcxf.maven.user.properties.prefixes to specify which user properties should be propagated to quarkus:create and quarkus:dev child processes.


ppalaga:
Thanks, @vkasala!

ppalaga:
@vkasala would you like to send a backport PR for 3.33?

PR #2120, opened 2026-04-30
[3.33] Stabilize DevModeTest

llowinge:


JiriOndrusek (approved):

integration-tests/server/src/test/java/io/quarkiverse/cxf/it/server/DevModeTest.java
```diff
@@ -239,8 +240,12 @@ static String[] args(String activeProfiles, Path settingsPath, String userProper
     }
 
     static void awaitResponse(String name, String description, String expectedDesciption) {
+        long minutes = ConfigProvider.getConfig()
+                .getOptionalValue("qcxf.test.devmodetest.await.timeout.minutes", Long.class)
+                .orElse(1L);
         Awaitility.await()
-                .atMost(30, TimeUnit.SECONDS)
+                .atMost(minutes, TimeUnit.MINUTES)
+                .pollInterval(5, TimeUnit.SECONDS)
```
ppalaga: @llowinge  why polling only every 5 seconds? It will prolong the test runtime duration by 5 seconds in the worst case. Isn't polling relatively cheap? If the default of 100 ms is too short, I'd vote for prolonging it to 200, 300 ms or the like.

llowinge: @ppalaga  Changed to 300ms, thanks.

PR #2121, opened 2026-04-30
Stabilize DevModeTest

llowinge:


JiriOndrusek (approved):
It would be probably a little bit safer to to use following code, but probably not necessary

```
long minutes;
try {
    minutes = Long.parseLong(System.getProperty("qcxf.test.devmodetest.await.timeout.minutes", "1").trim());
} catch (NumberFormatException e) {
    timeout = 1L;
}
```

instead of
```
long minutes = Long.parseLong(System.getProperty("qcxf.test.devmodetest.await.timeout.minutes", "1"));
```

llowinge:
@JiriOndrusek If user provides non-number value, i would expect it to fail and not silently use default. I'll try to use Quarkus config instead of `System.getProperty`.

JiriOndrusek:
> @JiriOndrusek If user provides non-number value, i would expect it to fail and not silently use default. I'll try to use Quarkus config instead of `System.getProperty`.

Sure, I thought that `.orElse(1L);` is  there as a default value, when there is no property

JiriOndrusek (approved):

integration-tests/server/src/test/java/io/quarkiverse/cxf/it/server/DevModeTest.java
```diff
@@ -239,8 +240,12 @@ static String[] args(String activeProfiles, Path settingsPath, String userProper
     }
 
     static void awaitResponse(String name, String description, String expectedDesciption) {
+        long minutes = ConfigProvider.getConfig()
+                .getOptionalValue("qcxf.test.devmodetest.await.timeout.minutes", Long.class)
+                .orElse(1L);
         Awaitility.await()
-                .atMost(30, TimeUnit.SECONDS)
+                .atMost(minutes, TimeUnit.MINUTES)
+                .pollInterval(5, TimeUnit.SECONDS)
```
ppalaga: Same as https://github.com/quarkiverse/quarkus-cxf/pull/2120#pullrequestreview-4216330468

llowinge: @ppalaga Changed to 300ms, thanks.

PR #2148, opened 2026-05-21
Revert "Temporary workaround for https://github.com/quarkiverse/quarkus-cxf/issues/1608"

### LINKED ISSUE #1608: [CXF] `TransportURIResolver.resolve()` swallows exceptions
`TransportURIResolver.resolve()` swallows exceptions around here:

https://github.com/apache/cxf/blob/cxf-4.0.5/core/src/main/java/org/apache/cxf/transport/TransportURIResolver.java#L100-L159

That makes it hard for us to fail if there is some blocking I/O (e.g. related to WSDL loading) on Vert.x event loop.

We should ask CXF folks what is the motivation for the swallowing and how could we best get around that at least when loading WSDLs.

ppalaga:
This reverts commit 1b1ce02f61ca614fbe151eec38ab1a62e07d76e9.

fix #1608


ppalaga:
This does not work well

PR #2193, opened 2026-06-15
[3.33] Parametrize also the Quarkus Platform groupId path

ppalaga:


docs/modules/ROOT/pages/user-guide/create-project.adoc
```diff
@@ -57,7 +57,7 @@ or other {link-quarkus-docs-base}/tooling#build-tool[Quarkus development tools],
     ...
     <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
     <quarkus.platform.group-id>{quarkus-platform-groupId}</quarkus.platform.group-id>
-    <quarkus.platform.version><!-- Check the latest {link-boms-maven-repo-base}/io/quarkus/platform/quarkus-cxf-bom/ --></quarkus.platform.version>
+    <quarkus.platform.version><!-- Check the latest {link-boms-maven-repo-base}/{quarkus-platform-groupId-path}/quarkus-cxf-bom/ --></quarkus.platform.version>
```
rhaetor, line 60:
```suggestion
    <quarkus.platform.version>{camel-quarkus-bom-version}<!-- Use the latest 
    {link-boms-maven-repo-base}/{quarkus-platform-groupId-path}/quarkus-cxf-bom/ -->
    </quarkus.platform.version>
```

rhaetor: How about this combination? 

Then you get a minimum version and also a recommendation to check.

ppalaga: There is no `camel-quarkus-bom-version` in Quarkus CXF antora config. Using `{quarkus-version}` would be along the same line, but there is no guarantee Quarkus Platform gets released with the same version number as Quarkus Core (which `{quarkus-version}` stores). Moreover, in the community, there is always a week delay between Quarkus Core and Platform. So during roughly one of four weeks the docs would be wrong. I think we should let it to the users to figure out the latest version. Anyway, we already encourage to use code.quarkus.io which uses the latest.

rhaetor: Very good. Thanks.

PR #2227, opened 2026-07-08
test: add wsdl2java -validate option tests

JinyuChen97:
Deployment unit tests call Wsdl2JavaCodeGen.wsdl2java() directly to verify that a WSDL with mixed binding style is accepted without -validate but rejected with -validate ("Mixed style, invalid WSDL").

The integration test verifies that a valid WSDL processed with -validate generates Java sources successfully end-to-end.

The negative case is not covered in the integration test because -validate runs at build time (generate-sources phase); a failing WSDL would crash the Maven build before any test code executes.

Co-authored-by: Bob <bob@ibm.com>

### DIFF
--- a/extensions/core/deployment/src/test/java/io/quarkiverse/cxf/deployment/codegen/Wsdl2JavaValidateTest.java
+++ b/extensions/core/deployment/src/test/java/io/quarkiverse/cxf/deployment/codegen/Wsdl2JavaValidateTest.java
@@ -1,0 +1,120 @@
+package io.quarkiverse.cxf.deployment.codegen;
+
+import static io.quarkiverse.cxf.deployment.codegen.Wsdl2JavaParamsTest.proxy;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import io.quarkiverse.cxf.deployment.CxfBuildTimeConfig.Wsdl2JavaParameterSet;
+
+public class Wsdl2JavaValidateTest {
+
+    @TempDir
+    Path tempDir;
+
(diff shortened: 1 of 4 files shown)


llowinge:
@JinyuChen97 Thank you very much. Very nice approach.

integration-tests/wsdl2java/src/test/java/io/quarkiverse/cxf/wsdl2java/it/Wsdl2JavaValidateTest.java
```diff
@@ -0,0 +1,17 @@
...
+
+import java.nio.file.Path;
+
+import org.junit.jupiter.api.Test;
+
+public class Wsdl2JavaValidateTest {
+
+    @Test
+    void validWsdlWithValidateSucceeds() {
+        assertThat(Path.of(
```
ppalaga: An equivalent assertion exists in io.quarkiverse.cxf.wsdl2java.it.Wsdl2JavaTest.codegenTests(). I'd vote for removing Wsdl2JavaValidateTest from this PR

extensions/core/deployment/src/test/java/io/quarkiverse/cxf/deployment/codegen/Wsdl2JavaValidateTest.java
```diff
@@ -0,0 +1,176 @@
...
+            return Path.of(getClass().getResource("/wsdllocation").toURI());
+        } catch (Exception e) {
+            throw new RuntimeException("Could not find test resources at /wsdllocation", e);
+        }
+    }
+
+    /**
+     * Minimal {@link Wsdl2JavaParameterSet} for use in tests.
+     */
+    static class TestParams implements Wsdl2JavaParameterSet {
```
ppalaga: Could you please use io.quarkiverse.cxf.deployment.codegen.Wsdl2JavaParamsTest.proxy(Object...) instead (make it public static if possible)

That approach is easier to maintain in the long term. With a dummy test implementation, one would have to update the test every time `Wsdl2JavaParameterSet` gets a new method added or removed.

ppalaga (changes requested):
Thanks, @JinyuChen97 the overall approach is good, I added some suggestions inline

ppalaga (approved):
Thanks, @JinyuChen97

PR #2239, opened 2026-07-28
Add Quarkus version mismatch check

### LINKED ISSUE #2111: Fail at built time if the Quarkus version is too different than expected
Sometimes, end users are not aware of the proper dependency management https://docs.quarkiverse.io/quarkus-cxf/dev/user-guide/create-project.html#dependency-management

We might implement a buildtime check of Quarkus version being used and fail the build (with some opt-out property in application.properties) if the Quarkus version is too far away from the expected one.

matheusandre1:
Closes: #2111

### DIFF
--- a/extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
+++ b/extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
@@ -31,6 +31,18 @@ public interface CxfBuildTimeConfig {
     Optional<List<String>> wsdlPath();
 
     /**
+     * What to do at build time when the `major.minor` version of Quarkus used by the application differs from the
+     * Quarkus version this version of Quarkus CXF was built with. Check the
+     * xref:user-guide/create-project.adoc#dependency-management[Dependency management] section of the user guide for
+     * how to fix such a mismatch.
+     *
+     * @asciidoclet
+     * @since 3.39.0
+     */
+    @WithDefault("fail")
+    ErrorRemedy onQuarkusVersionMismatch();
+
+    /**
      * Build time configuration options for Quarkus code generation phase.
      *
      * @asciidoclet
--- a/docs/modules/ROOT/pages/reference/extensions/quarkus-cxf.adoc
+++ b/docs/modules/ROOT/pages/reference/extensions/quarkus-cxf.adoc
@@ -82,6 +82,18 @@ icon:lock[title=Fixed at build time] Configuration property fixed at build time.
 | Configuration property | Type | Default
 
 
+.<|icon:lock[title=Fixed at build time] [[quarkus-cxf_quarkus-cxf-on-quarkus-version-mismatch]]`link:#quarkus-cxf_quarkus-cxf-on-quarkus-version-mismatch[quarkus.cxf.on-quarkus-version-mismatch]`
+.<| tooltip:fail[Fail the build], tooltip:warn[Log a warning and continue the build], tooltip:ignore[Do not perform the check at all]
+.<| `fail`
+
+3+a|What to do at build time when the `major.minor` version of Quarkus used by the application differs from the
+Quarkus version this version of Quarkus CXF was built with. Check the
+xref:user-guide/create-project.adoc#dependency-management[Dependency management] section of the user guide for
+how to fix such a mismatch.
+
+*Environment variable*: `+++QUARKUS_CXF_ON_QUARKUS_VERSION_MISMATCH+++` +
+*Since Quarkus CXF*: 3.39.0
+
 .<|icon:lock[title=Fixed at build time] [[quarkus-cxf_quarkus-cxf-codegen-wsdl2java-enabled]]`link:#quarkus-cxf_quarkus-cxf-codegen-wsdl2java-enabled[quarkus.cxf.codegen.wsdl2java.enabled]`
 .<| `boolean`
 .<| `true`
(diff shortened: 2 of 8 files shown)


extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -45,6 +68,21 @@ public interface CxfBuildTimeConfig {
...
+        FAIL,
+        /**
+         * Log a warning and continue the build
+         */
+        WARN,
+        /**
+         * Do not perform the check at all
+         */
+        IGNORE
+    }
```
ppalaga: Could you please move the enum to a separate file and name it `ErrorRemedy`?

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -30,6 +30,29 @@ public interface CxfBuildTimeConfig {
...
+     * - `warn` - log a warning and continue the build
+     * - `ignore` - do not perform the check at all
+     *
+     * Quarkus CXF is tested only against the Quarkus version it was built with. Combining it with any other Quarkus
+     * version stream is not supported and may lead to obscure build time or runtime failures. Check the
+     * xref:user-guide/create-project.adoc#dependency-management[Dependency management] section of the user guide for
+     * more information about how to select mutually compatible versions of Quarkus and Quarkus CXF.
+     *
+     * @asciidoclet
+     * @since 3.38.1
```
ppalaga:
```suggestion
     * @since 3.39.0
```

We won't publish new features in a micro release.

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -30,6 +30,29 @@ public interface CxfBuildTimeConfig {
...
+     * The possible values are as follows:
+     *
+     * - `fail` - fail the build
+     * - `warn` - log a warning and continue the build
+     * - `ignore` - do not perform the check at all
+     *
+     * Quarkus CXF is tested only against the Quarkus version it was built with. Combining it with any other Quarkus
+     * version stream is not supported and may lead to obscure build time or runtime failures. Check the
+     * xref:user-guide/create-project.adoc#dependency-management[Dependency management] section of the user guide for
+     * more information about how to select mutually compatible versions of Quarkus and Quarkus CXF.
```
ppalaga: Too much text, too much details for what the option does. Please reduce.

extensions/core/deployment/src/main/java/io/quarkiverse/cxf/deployment/CxfBuildTimeConfig.java
```diff
@@ -30,6 +30,29 @@ public interface CxfBuildTimeConfig {
...
+     *
+     * @asciidoclet
+     * @since 3.38.1
+     */
+    @WithDefault("fail")
+    QuarkusVersionMismatchAction quarkusVersionMismatch();
```
ppalaga:
```suggestion
    QuarkusVersionMismatchAction onQuarkusVersionMismatch();
```

ppalaga (commented):
Thanks for the PR, @matheusandre1! I added some comments inline, but when looking at it overall, there is not much Quarkus CXF specific in this PR. 

@gsmet I wonder whether we should have extension compatibility checking in Quarkus Core? It does not look like we use the `requires-quarkus-core` metadata for this kind of purpose? Would that be a good idea and may  @matheusandre1 send a PR to Quarkus?

matheusandre1:
> Thanks for the PR, @matheusandre1! I added some comments inline, but when looking at it overall, there is not much Quarkus CXF specific in this PR.
> 
> @gsmet I wonder whether we should have extension compatibility checking in Quarkus Core? It does not look like we use the `requires-quarkus-core` metadata for this kind of purpose? Would that be a good idea and may @matheusandre1 send a PR to Quarkus?

I'm going to work up the courage to propose a PR for this in the core.

ppalaga:
@matheusandre1 I'd wait for @gsmet to express his opinion whether it is a good idea.

ppalaga:
I spoke with Quarkus team. We agreed, that we want the validation of `requires-quarkus-core` metadata in Quarkus Core at build time. They confirmed there is no such check yet.

Initially, I suggest to implement only the enforcement of `requires-quarkus-core`. 

With its current default value `[3.38,)` for Quarkus CXF and other extension it would only prevent running with too old Quarkus Core, but would not stop running with a too new Quarkus Core. I think that's fine and I think we should come up with a way to set the proper version upper bound in Quarkus CXF's `quarkus-extension.yaml`.

ppalaga:
> we should come up with a way to set the proper version upper bound in Quarkus CXF's quarkus-extension.yaml.

Given that we sometimes miss a minor Quarkus Platform release, we should aim at having `requires-quarkus-core: "[current-major.current-minor,current-major.(current-minor+2))"` in Quarkus CXF's quarkus-extension.yaml.

ppalaga:
@matheusandre1 would you like to generalize this PR for all extensions available in the application being built and move it to https://github.com/quarkusio/quarkus ?

matheusandre1:
I will create the issue in the core and implement it.

ppalaga:
We agreed, this will be done in Quarkus.

PR #2272, opened 2026-08-27
Change human readable names of XJC extensions, explain XJC

ppalaga:


rolfedh (approved):
LGTM

rolfedh:
Two small consistency points I missed on my first read, both cosmetic.

**Capitalization of "plugins" differs between the two extensions.** `quarkus-cxf-xjc-plugins` uses "XJC **P**lugins" and `quarkus-cxf-jaxb-plugins` uses "XJC **p**lugins", so the two names no longer match each other:

- `XJC Plugins from Apache CXF project for wsdl2java`
- `XJC plugins from org.jvnet.jaxb:jaxb-plugins for wsdl2java`

The xjc-plugins extension is also inconsistent with itself: the page H1 and the extension name use "Plugins", while `intro.adoc` uses "plugins".

**"from Apache CXF project" reads better as "from the Apache CXF project".** It appears without the article in 9 added lines, all on the xjc-plugins side: `nav.adoc`, the page H1, `intro.adoc` (twice, since the reference page includes it), `reference/index.adoc`, the three `pom.xml` files, and `quarkus-extension.yaml`.

Neither is worth blocking on. Flagging them because the `quarkus-extension.yaml` `name` and the Maven `<name>` values surface in the extension catalog and on the project generators, so they get more visibility than a docs heading.

Happy to push a commit if that is easier than doing it yourself.

Thanks for turning the discussion around so quickly, and for adding the XJC explanation note. That was the part our readers were missing.

ppalaga:
Thanks for the feedback. I changed it everywhere to lower case plugins.

rolfedh (approved):

PR #2285, opened 2026-09-08
Override CXF messages related to addressing.decoupled properly

### LINKED ISSUE #2283: NPE in replaceCxfCoreResourceBundles when the application re-augments (mutable-jar + quarkus.launch.rebuild)
### Describe the bug

`QuarkusCxfProcessor#replaceCxfCoreResourceBundles`, which is new in 3.39.0, throws an NPE when the application is **re-augmented** rather than augmented as part of a Maven build:

```
Build step io.quarkiverse.cxf.deployment.QuarkusCxfProcessor#replaceCxfCoreResourceBundles
threw an exception: java.lang.NullPointerException: inStream parameter is null
    at java.base/java.util.Objects.requireNonNull(Objects.java:246)
    at java.base/java.util.Properties.load(Properties.java:409)
    at io.quarkiverse.cxf.deployment.QuarkusCxfProcessor.replaceCxfCoreResourceBundles(QuarkusCxfProcessor.java:236)
    at java.base/java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:735)
    at io.quarkus.deployment.ExtensionLoader$3.execute(ExtensionLoader.java:811)
    at io.quarkus.builder.BuildContext.run(BuildContext.java:242)
    at org.jboss.threads.ContextHandler$1.runWith(ContextHandler.java:18)
    at org.jboss.threads.EnhancedQueueExecutor$Task.doRunWith(EnhancedQueueExecutor.java:2651)
```

The build step resolves the resource bundle like this:

```java
InputStream in = ContextUtils.class.getClassLoader()
        .getResourceAsStream("org/apache/cxf/ws/addressing/Messages.properties");
props.load(in);
```

`getResourceAsStream` returns `null` under the re-augmentation classloader setup, and the result is handed to `Properties.load` unchecked.

This is disruptive because re-augmentation is the documented way to change build-time configuration without rebuilding ([guide](https://quarkus.io/guides/reaugmentation)). Our container images are packaged as `mutable-jar` and start with `-Dquarkus.launch.rebuild=true`, so **every boot re-augments** — on 3.39.x the application no longer starts at all.

Worth highlighting: a plain `mvn package` / `mvn verify` succeeds on the affected versions, because that only exercises build-time augmentation. A normal CI pipeline therefore stays green while the produced image is unable to boot.

### Expected behavior

Re-augmentation succeeds, as it does on 3.37.0 and 3.38.0.

### Actual behavior

Augmentation aborts with the NPE above and the application does not start.

### How to Reproduce?

1. Any application using `quarkus-cxf` 3.39.0 or 3.39.1.
2. Set `quarkus.package.jar.type=mutable-jar` in `application.properties`.
3. `mvn package -DskipTests` — succeeds.
4. `java -Dquarkus.launch.rebuild=true -jar target/quarkus-app/quarkus-run.jar` — fails with the NPE.

Bisected on our application (a SOAP client using `quarkus-cxf`), varying only `quarkus-cxf.version`:

| quarkus-cxf | `mvn package` | re-augment |
|---|---|---|
| 3.37.0 | ok | ok (build step does not exist yet) |
| 3.38.0 | ok | ok |
| 3.39.0 | ok | **NPE** |
| 3.39.1 | ok | **NPE** |

`replaceCxfCoreResourceBundles` is absent from `QuarkusCxfProcessor` at tags 3.37.0 and 3.38.0, and present at 3.39.0 and 3.39.1.

### Output of `uname -a` or `ver`

`Darwin 25.6.0` (arm64). Originally observed in a Linux container image, then reproduced locally.

### Output of `java -version`

```
openjdk version "25.0.2" 2026-01-20 LTS
OpenJDK Runtime Environment Temurin-25.0.2+10 (build 25.0.2+10-LTS)
```

### Quarkus version or git rev

Quarkus platform 3.38.x / 3.39.x, quarkus-cxf 3.39.0 and 3.39.1.

### Additional information

Either of these would fix it independently:

1. Null-check the stream and skip the override when the resource is not visible, optionally logging at DEBUG. The overrides are cosmetic message text, so degrading gracefully seems acceptable.
2. Resolve the resource from a classloader that is correct at augmentation time — e.g. the deployment classloader available to the build step — rather than `ContextUtils.class.getClassLoader()`.

Happy to test a snapshot against our re-augmenting setup if that helps. We are pinned to 3.38.0 in the meantime.

ppalaga:
fix #2283

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -160,7 +160,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        testModule: ['client', 'client-server', 'fastinfoset', 'metrics', 'mtls', 'mtls -Djks', 'mtom-awt', 'opentelemetry', 'saaj', 'santuario-xmlsec', 'server', 'ws-rm-client', 'ws-security', 'ws-security -Djks', 'ws-security-policy', 'ws-security-policy -Djks', 'ws-trust', 'wsdl2java', 'wsdl2java-no-config']
+        testModule: ['client', 'client-server', 'fastinfoset', 'metrics', 'mtls', 'mtls -Djks', 'mtom-awt', 'mutable-jar', 'opentelemetry', 'saaj', 'santuario-xmlsec', 'server', 'ws-rm-client', 'ws-security', 'ws-security -Djks', 'ws-security-policy', 'ws-security-policy -Djks', 'ws-trust', 'wsdl2java', 'wsdl2java-no-config']
     name: ${{matrix.testModule}} native tests
     needs: build-and-run-jvm-tests
     runs-on: ubuntu-latest
--- a/bom-test/pom.xml
+++ b/bom-test/pom.xml
@@ -51,7 +51,12 @@
             </dependency>
             <dependency>
                 <groupId>org.cli-assured</groupId>
-                <artifactId>cli-assured-mvn</artifactId>
+                <artifactId>cli-assured</artifactId>
+                <version>${cli-assured.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.cli-assured</groupId>
+                <artifactId>cli-assured-maven</artifactId>
                 <version>${cli-assured.version}</version>
             </dependency>
             <dependency>
--- a/docs/antora.yml
+++ b/docs/antora.yml
@@ -8,7 +8,7 @@ asciidoc:
   attributes:
 
     # Versions
-    quarkus-version: 3.39.0 # replace ${quarkus.version}
+    quarkus-version: 3.39.2 # replace ${quarkus.version}
     quarkus-cxf-version: 3.39.1 # replace ${release.current-version}
 
     # Toggle whether some page elements are rendered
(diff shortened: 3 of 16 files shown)


narcismatei:
Tested, and it fixes the problem for us. Thanks for the quick turnaround, and for adding `MutableJarLaunchRebuildIT` — that boot path had no coverage, which is exactly why this reached us in production.

**Setup:** a SOAP client application packaged as `quarkus.package.jar.type=mutable-jar`, started with `-Dquarkus.launch.rebuild=true` so it re-augments on every container boot. Quarkus 3.39.2, Java 25, CXF 4.2.3 (unchanged between the broken and fixed builds, so the messages-override rewrite is the only delta).

**Results**, re-augmenting repeatedly and counting exit codes:

| quarkus-cxf | re-augment runs |
|---|---|
| 3.39.1 | 0 pass / 20 fail |
| this PR @ f14bd3c1 | 20 pass / 0 fail |
| this PR @ 324fb92b | 10 pass / 0 fail |

Our full application test suite (208 tests) is also green against the PR build.

**One correction to my original report**, which may be useful to you: I described the failure as consistent, and on a normal multi-core machine it is — 20/20 on 3.39.1. But in our cluster, where the pod is limited to **1 CPU**, the same image succeeded on the **12th** boot after 11 identical failures. Same digest, same config, same feature flags. So `getResourceAsStream` returning null was timing-dependent, not absolute, which I think supports your reading that resource overriding was never reliably in place. The bytecode transformation approach removes that whole class of problem rather than narrowing the window.

**Not verified on our side:** that the overridden `addressing.decoupled` message text actually renders. That needs a rejected decoupled destination, which our application does not exercise — I only confirmed the NPE is gone and nothing regressed.

ppalaga:
Nice to hear that, thanks for the verification @narcismatei

PR #2295, opened 2026-09-17
Parametrize links to simplify building downstream documentation

rolfedh:
Gives the Prerequisites and Building a native executable links on the Create a new project page their own attributes, `link-quarkus-docs-prerequisites` and `link-quarkus-docs-building-native-image`. A downstream documentation build can then override each link with one value instead of patching the page, which `link-quarkus-docs-base` alone cannot do when the downstream URLs do not share the quarkus.io slugs. Both attributes resolve to the current URLs, so the community site renders unchanged.


ppalaga (approved):
Thanks @rolfedh!

Again, the PR description contained too much text for what the PR did.

PR #2297, opened 2026-09-18
Refer to the Generate your application button by its label, not its color

rolfedh:
Naming a control only by its color is an accessibility problem. The step now names the button by its label, in bold as a UI element.


rolfedh:
@ppalaga

PR #2299, opened 2026-09-18
Give the Caller threads and callback threads section an explicit ID

rolfedh:
The `quarkus.cxf.client.worker-dispatch-timeout` description linked this section by its generated ID, which changes with the `idprefix` and `idseparator` settings, so the link breaks in downstream documentation builds that use other values. An explicit ID, like the other linked sections on the page have, keeps it working everywhere.


rolfedh:
@ppalaga

